PR #52, opened 2026-04-08
cmake: better use shared libraries, parallelize build

twitchyliquid64:
cmake is a slow package, speeding it up basically directly speeds up build time.


twitchyliquid64:
```
package: cmake
spec name matches dir...Pass
spec name valid...Pass
cycle breaker naming...Pass
output naming...Pass
enumerate bins...Pass
output types valid...Pass
missing runtime_deps...Pass
build scripts are executable...Pass
build script disallowed-patterns...Pass
standalone tests...Pass
parse...Pass
import line...Pass
adjacent import...Pass
fmt...Pass
```

msample (approved):
LGTM - nice find!

PR #55, opened 2026-04-09
Improve packages which represent python dependencies

twitchyliquid64:
All python packages are unambiguously named as such


msample (approved):
LGTM

PR #72, opened 2026-04-15
Add node-lts package (Node.js v24.14.1 LTS "Krypton")

mitodrummer:
## Summary

- Adds a new `node-lts` package for Node.js v24.14.1 LTS ("Krypton")
- Includes `build.ncl` with the build spec and `build.sh` build script
- Node.js binary sourced from the official release tarball

## Test plan

- [ ] `min check --packages node-lts` passes
- [ ] `min patched-pkg node-lts` builds successfully
- [ ] `node --version` outputs `v24.14.1`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/node-lts/build.ncl
+++ b/packages/node-lts/build.ncl
@@ -35,6 +35,8 @@ let version = "24.14.1" in
     toolchain,
   ],
   runtime_deps = [
+    # Note: lief is absent here (unlike packages/node/) because --shared-lief was not added
+    # to Node.js configure until v25; the v24 LTS branch does not support it.
     coreutils, # npx/npm need /usr/bin/env
     openssl,
     c-ares,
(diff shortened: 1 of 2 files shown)


bryan-minimal (commented):
Welcome to the repo! Nice first PR — good instinct to copy-and-adapt from `packages/node/` rather than writing from scratch, and I noticed you correctly updated the `tar -xof` and `cd` paths in build.sh to match the new `node-v${V}` filename format (easy to miss when copying).

One question worth a comment: **`lief` is dropped from runtime_deps and `--shared-lief` from configure, consistently in both files.** Existing `packages/node/` has both. Is that because Node 24 LTS doesn't support `--shared-lief` yet, or because it's not needed for the LTS stream? Either is fine — just worth a one-line comment in `runtime_deps` so future maintainers know why the LTS variant diverges.

On the direct `nodejs.org` URL vs the GCS-mirrored URL that `packages/node/` uses: we do want everything mirrored to GCS eventually for resilience against upstream outages, but that's a broad repo-wide effort and not for this PR to foist on you. Direct upstream is fine here.

msample (dismissed):
LGTM.  First package! 🚀

msample (approved):
LGTM

PR #86, opened 2026-04-18
Quick update batch

twitchyliquid64:
- [x] Built successfully by buildbot, see run `1fda4978-efd9-4526-a4c8-1c83a4674545`
 - [x] `minimal check` succeeds

### DIFF
--- a/packages/file/build.ncl
+++ b/packages/file/build.ncl
@@ -12,14 +12,16 @@ let toolchain = import "../toolchain/build.ncl" in
 let glibc = import "../glibc/build.ncl" in
 let lz4 = import "../lz4/build.ncl" in
 let xz = import "../xz/build.ncl" in
-let version = "5.46" in
+let version = "5.47" in
 {
   name = "file",
   build_deps = [
     { file = "build.sh" } | Local,
     {
+      # Home page: https://www.darwinsys.com/file/
+      # https://astron.com/pub/file/file-5.47.tar.gz
       url = "gs://minimal-staging-archives/file-%{version}.tar.gz",
-      sha256 = "c9cc77c7c560c543135edc555af609d5619dbef011997e988ce40a3d75d86088"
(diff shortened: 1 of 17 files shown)


packages/file/build.ncl
```diff
@@ -12,14 +12,16 @@ let toolchain = import "../toolchain/build.ncl" in
...
 let xz = import "../xz/build.ncl" in
-let version = "5.46" in
+let version = "5.47" in
 {
   name = "file",
   build_deps = [
     { file = "build.sh" } | Local,
     {
+      # Home page: https://www.darwinsys.com/file/
+      # https://astron.com/pub/file/file-5.47.tar.gz
```
msample, line 22: do we support a source provenance of URL?  Better there than a comment?

packages/libcap/build.ncl
```diff
@@ -10,14 +10,16 @@ let coreutils = import "../coreutils/build.ncl" in
...
 {
   name = "libcap",
   build_deps = [
     { file = "build.sh" } | Local,
     {
-      url = "gs://minimal-staging-archives/libcap-%{version}.tar.gz",
-      sha256 = "5e339f5ab434bf5b7979f65437ca24942ddcd6e6a3e57347370cd791bc0ea51c",
+      # Home page: https://sites.google.com/site/fullycapable/
+      # Download: https://www.kernel.org/pub/linux/libs/security/linux-privs/libcap2/libcap-2.78.tar.xz
+      url = "gs://minimal-staging-archives/libcap-%{version}.tar.xz",
```
msample, line 21: same as other comment about source provenance URL as data vs comment.

packages/file/build.ncl
twitchyliquid64: We do not unfortunately. We should add that, but it will be after this PR.

packages/libcap/build.ncl
twitchyliquid64: As above: not implemented today but we should, comment for now.

msample (approved):

PR #97, opened 2026-04-22
pnpm: use node-lts runtime, frozen-lockfile installs

mitodrummer:
## Summary
- Switch the pnpm harness's `runtime_packages` from `node` to `node-lts`.
- Switch the pnpm package's `runtime_deps` from `node` to `node-lts` to match.
- In the harness install step, use `pnpm install --frozen-lockfile` when `pnpm-lock.yaml` is present, falling back to plain `pnpm install` otherwise.

## Reasoning
**node-lts:** pnpm's own releases are built against Node LTS, and most hosting providers default to LTS for Node runtimes. Aligning both the harness runtime and the pnpm package's declared runtime with LTS matches what pnpm is tested against upstream and what users are most likely to deploy on.

**--frozen-lockfile:** Mirrors the npm harness's `npm ci` behavior — when a lockfile is checked in, installs should be reproducible and fail loudly if the lockfile drifts from `package.json`. Falls back to plain `pnpm install` when no lockfile is present (e.g. a workspace root matched only by `pnpm-workspace.yaml`).

## Test plan
- [ ] `min check --harnesses pnpm`
- [ ] `min check --packages pnpm`

### DIFF
--- a/harnesses/pnpm/harness.ncl
+++ b/harnesses/pnpm/harness.ncl
@@ -10,7 +10,9 @@ harness {
     "-c",
     m%"
     # Run pnpm install if no packages are installed (first start)
-    [ -z "$(pnpm list 2>/dev/null)" ] && echo "/bin/pnpm install"
+    if [ -z "$(pnpm list 2>/dev/null)" ]; then
+      [ -f pnpm-lock.yaml ] && echo '/bin/pnpm install --frozen-lockfile' || echo '/bin/pnpm install'
+    fi
     echo '/bin/pnpm build'
   "%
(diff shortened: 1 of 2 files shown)


twitchyliquid64 (commented):
I think we also need to update the `pnpm` package, which declares a runtime dependency on `node` explicitly:

https://github.com/gominimal/pkgs/blob/eac36d402027254327e492e9b531302b38dc4148/packages/pnpm/build.ncl#L21

twitchyliquid64 (approved):

PR #101, opened 2026-04-22
Add opencode package

norrietaylor:
## Summary
- Adds `packages/opencode/` packaging the [anomalyco/opencode](https://github.com/anomalyco/opencode) CLI at v1.14.20 (formerly `sst/opencode` — upstream repo was transferred).
- Builds from source via `bun` (rather than repackaging the prebuilt release tarball), matching the policy now documented in `CLAUDE.md`.
- Unpacks the source tarball manually because the built-in extractor trips on bracketed filenames in the monorepo, relaxes the `packageManager` pin to whatever `bun` version pkgs ships, then runs `script/build.ts --single` for a native single-target binary.
- Declares `bun` as a build dep and `glibc` / `git` / `ripgrep` as runtime deps. Wires `~/.local/share/opencode` and `~/.config/opencode` as state dirs and retains the `opencode --version` smoketest.
- Adds an `opencode` task to `minimal.toml` so the CLI can be launched via `minimal run opencode`.
- Updates `CLAUDE.md` with two new authoring rules: verify the canonical upstream repo (catches silent GitHub redirects like `sst → anomalyco`) and prefer building from source when the required toolchain is already packaged.

## Test plan
- [x] `min check --packages opencode` — all checkers pass
- [x] `min patched-pkg opencode` — builds from source successfully
- [x] `opencode --version` smoketest passes post-build

### DIFF
--- a/packages/opencode/build.ncl
+++ b/packages/opencode/build.ncl
@@ -1,35 +1,34 @@
-let { standaloneTest, Attrs, BuildSpec, Local, OutputBin, Source, .. } = import "minimal.ncl" in
-let { target, .. } = import "config.ncl" in
+let { standaloneTest, Attrs, BuildSpec, Local, Needs, OutputBin, Source, .. } = import "minimal.ncl" in
 let base = import "../base/build.ncl" in
+let bun = import "../bun/build.ncl" in
 let git = import "../git/build.ncl" in
 let glibc = import "../glibc/build.ncl" in
+let ripgrep = import "../ripgrep/build.ncl" in
 
 let version = "1.14.20" in
 {
   name = "opencode",
   build_deps = [
     { file = "build.sh" } | Local,
-    match {
-      { arch = 'Amd64, .. } =>
-        {
-          url = "https://github.com/sst/opencode/releases/download/v%{version}/opencode-linux-x64.tar.gz",
-          sha256 = "1707133022382ea8c8cc5bc47ab1db52fa206bd9baf8d3ad4d862b489dbb448f",
-          extract = true,
-        } | Source,
-      { arch = 'Arm64, .. } =>
-        {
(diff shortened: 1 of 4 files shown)


packages/opencode/build.ncl
```diff
@@ -0,0 +1,60 @@
...
+{
+  name = "opencode",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    match {
+      { arch = 'Amd64, .. } =>
+        {
+          url = "https://github.com/sst/opencode/releases/download/v%{version}/opencode-linux-x64.tar.gz",
+          sha256 = "1707133022382ea8c8cc5bc47ab1db52fa206bd9baf8d3ad4d862b489dbb448f",
+          extract = true,
```
msample: isn't "https://github.com/anomalyco/opencode" the updated repo for this?

packages/opencode/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+set -e
+
+mkdir -p "$OUTPUT_DIR/usr/bin"
+install -m 755 opencode "$OUTPUT_DIR/usr/bin/opencode"
```
msample: ideally we build from source where possible, esp if the required toolchain is already in pkgs.  This allows us to tweak build scripts for reproducibility in some cases, and it also eliminates concerns about mutated release artifacts.

norrietaylor: 🤦 I thought this worked a little too magically. Thanks for the catch!

packages/opencode/build.ncl
norrietaylor: Confirmed: sst/opencode redirects to anomalyco/opencode.

twitchyliquid64 (approved):

PR #102, opened 2026-04-22
CLAUDE.md: expand harness/needs/prebuilt sections, fix typos

norrietaylor:
## Summary
- Add a Harnesses overview with an example and the list of current harnesses
- Document the previously-undescribed `needs` and `prebuilt` fields
- Replace the short tar workaround note with a concrete manual-extraction example
- Fix missing apostrophes (`arent`, `theres`, `hasnt`)

## Test plan
- [x] Render CLAUDE.md and verify formatting

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,3 +1,7 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
 # pkgs repository
 
 You are in the packaging repository for the Minimal build system & package manager, containing the
@@ -9,7 +13,7 @@ You are running in an environment with access to a few specialized tools to help
 goals. These tools are all subcommands of the `min` command.
  
  * `min add <package name>` - Installs the package with the given name into your environment, letting you
-    use the CLI tools it encapsulates. If you arent sure of the right package name, you can use
+    use the CLI tools it encapsulates. If you aren't sure of the right package name, you can use
     `min search <term>` to help find it.
  * `min check [--packages] [--profiles] [--harnesses] [--fix] [<name 1>[, <name N>]]` - Runs linters and static checks on the packages/profiles/harnesses with the given names, or all if names are not specified. If none of --profiles, --harnesses, and --packages are set, then checks are run for all three object kinds.
  * `min patched-pkg <package name>` - Runs the build for the named package. Unlike a full build, a patched build will wire dependencies to the most recent available version of the package with the same name, so you won't have long rebuilds when editing packages which are circularly dependent on a lot of other packages.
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -274,7 +313,20 @@ Omit it if its not one of these variants.
...
+```ncl
+needs = {
+  dns = {},
+  internet = {},
+},
+```
+
+#### `prebuilt`
+
+Used for packages whose artifacts are downloaded pre-compiled rather than built from source. Not covered in detail here — check existing packages that use it for reference.
```
bryan-minimal: Nice to surface this field — it's been under-documented. Small suggestion: "check existing packages that use it for reference" tells the reader *where to look* without first telling them *what it means*, which is a bit of an anti-pattern for docs. One extra line on the semantics would make the section self-contained:

```suggestion
Declares the package's output as a checked-in prebuilt binary rather than something built from source during the pipeline. Typically used for toolchain-bootstrap packages that need a working binary before the toolchain itself can compile anything — see `packages/bash-bootstrap/build.ncl` as an example.
```

Not blocking — just a style nudge.

mitodrummer (approved):

CLAUDE.md
```diff
@@ -24,7 +28,28 @@ Specifically:
 
  * Packages: `packages/<package name>/build.ncl`
  * Harnesses: `harnesses/<harness name>/harness.ncl`
- * Profile: `profiles/<profile name>/profile.ncl`
+
+The repo-level config lives at `minimal.toml` (declares the minimum `stdlib` version and interactive `tasks` like `min run shell` / `min run claude`).
+
+### Harnesses
+
+A harness describes a reusable build environment for a class of project (e.g. a Go module, a Rust crate, a CMake project). Each harness declares the packages it needs, a default build command, and a set of project-detection rules — `min` uses these rules to auto-select the right harness for a source tree.
```
twitchyliquid64: s/ `min` / `minimal init`

CLAUDE.md
```diff
@@ -208,7 +233,21 @@ of the build. You can also have the system unpack the tarball into the working d
...
+  { file = "build.sh" } | Local,
+  { url = "...", sha256 = "..." } | Source,
+  tar,
+  ...
+],
+```
+
+```bash
+tar -xf source-tarball.tar.gz
+cd source-dir
```
twitchyliquid64: We may want to mention `build_args` and passing in the version from the `build.ncl`, so that `build.sh` doesnt need to change on every version bump. EG: 

https://github.com/gominimal/pkgs/blob/caf166c922a11c5b0145f1acf15360db02178190/packages/bash/build.ncl#L45-L47

https://github.com/gominimal/pkgs/blob/caf166c922a11c5b0145f1acf15360db02178190/packages/bash/build.sh#L4

CLAUDE.md
```diff
@@ -274,7 +313,20 @@ Omit it if its not one of these variants.
...
+```ncl
+needs = {
+  dns = {},
+  internet = {},
+},
+```
+
+#### `prebuilt`
+
+Declares the package's output as a checked-in prebuilt binary rather than something built from source during the pipeline. Typically used for toolchain-bootstrap packages that need a working binary before the toolchain itself can compile anything — see `packages/bash-bootstrap/build.ncl` as an example.
```
twitchyliquid64, line 350: Also worth noting that the directory layout of a tarball which is the source of a `prebuilt=true` package must match the layout on disk of files that package is emitting. EG: a prebuilt `bash` package would need a tarball with the bash binary at `usr/bin/bash` - theres no opportunity to move things around.

PR #107, opened 2026-04-22
Update bun to 1.3.13

jtnkminimal:
Switch the build to bun's own build:release orchestration (handles codegen, cmake deps, zig, link, and strip) and add a python-backed unzip shim so the nested zig toolchain extraction works in the minimal sandbox.

### DIFF
--- a/packages/bun/build.sh
+++ b/packages/bun/build.sh
@@ -1,44 +1,12 @@
 #!/bin/sh
 set -ex
 
-# Provide an `unzip` shim backed by python's zipfile module (no unzip package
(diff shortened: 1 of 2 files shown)


packages/bun/build.sh
```diff
@@ -1,6 +1,38 @@
 #!/bin/sh
 set -ex
 
+# Provide an `unzip` shim backed by python's zipfile module (no unzip package
```
mitodrummer: did we want to add 'unzip' as a min package?

jtnkminimal: yea good idea. i'll refactor!

mitodrummer (approved):

PR #133, opened 2026-04-29
llvm: restrict targets to host+embedded, cap link parallelism, template version

bryan-minimal:
## What

Three independent LLVM build improvements:

| Change | Effect |
|---|---|
| Restrict `LLVM_TARGETS_TO_BUILD` from `"all"` (~14 backends) to `"X86;AArch64;WebAssembly;ARM;RISCV"` | ~50% LLVM compile-time reduction |
| Add `LLVM_PARALLEL_LINK_JOBS=2` | Caps link-phase RAM, same OOM fix Tom recently applied to mono and or-tools |
| Add `build_args = { include version }` + version-template `build.sh` | Future LLVM bumps are a single edit in `build.ncl` (CR-recommended pattern) |

## Targets restriction — verification done

| Backend dropped | Used by anything in pkgs? |
|---|---|
| PowerPC, MIPS, SPARC, Hexagon | No |
| AMDGPU, NVPTX | No |
| AVR, BPF, LoongArch, M68k, MSP430, SystemZ, VE, XCore | No |

Embedded ARM cross-compile uses dedicated `gcc-arm-none-eabi` + `binutils-arm-none-eabi` packages (GNU toolchain). They don't touch LLVM's ARM backend.

Kept ARM + RISCV anyway because zig (which links shared LLVM via `ZIG_SHARED_LLVM=ON`) advertises cross-compile-anywhere — preserving the most plausible downstream use cases.

## What's NOT in this PR (deferred)

Originally tried to add `libcxx`, `libcxxabi`, `openmp` via `LLVM_ENABLE_RUNTIMES`. The standalone per-component tarballs don't reassemble into a layout that satisfies LLVM 21's `runtimes/CMakeLists.txt` path expectations — the orchestrator hardcodes `${CMAKE_CURRENT_SOURCE_DIR}/../../<name>` AND spawns a child cmake build via `llvm_ExternalProject_Add(runtimes <self>/../../runtimes)` which expects a second orchestrator copy. In the monorepo both files exist as distinct paths; in the per-component tarballs only one ships and there's no clean way to satisfy both expectations.

Three layout attempts (PROJECTS, RUNTIMES at top-level, RUNTIMES under `<llvm-src>/runtimes/`) failed in different ways. Right vehicle is a focused follow-up using the **monorepo tarball** (`llvm-project-21.1.8.src.tar.xz`), which has the layout cmake expects out of the box AND collapses our 8 source tarballs to 1 — incidentally resolving [pkgmgr-rs#59](https://github.com/gominimal/pkgmgr-rs/issues/59) (multi-source updater gap) for LLVM. Will file a follow-up issue.

Also deliberately not in this PR:
- **lldb** — would add ~20-30% build time + needs new `swig` + `libedit` packages we don't have. No current debugger demand.
- **mlir, polly, flang, bolt** — specialty subprojects, no current demand.

## Validation

Diff is small + targeted. Real validation is "the buildbot rebuilds LLVM successfully" — this is a cache-miss change (new spec hash), so the buildbot will exercise the full build path.

### DIFF
--- a/packages/llvm/build.ncl
+++ b/packages/llvm/build.ncl
@@ -76,11 +76,14 @@ let version = "21.1.8" in
   ],
 
   cmd = "./build.sh",
+  build_args = { include version },
   outputs = {
     bins = { glob = "usr/bin/*" } | OutputBin,
     clangd = { glob = "usr/bin/clangd" } | OutputBin,
 
     libllvm = { glob = "usr/lib/*LLVM*.{so*,a}" } | OutputLib,
+    libcxx_runtime = { glob = "usr/lib/libc++*.{so*,a}" } | OutputLib,
+    libomp = { glob = "usr/lib/libomp*.{so*,a}" } | OutputLib,
     libear = { glob = "usr/lib/libear/**", allow_data = true } | OutputLib,
     liblto = { glob = "usr/lib/libLTO.so*" } | OutputLib,
     libremarks = { glob = "usr/lib/libRemarks.so*" } | OutputLib,
--- a/packages/llvm/build.ncl
+++ b/packages/llvm/build.ncl
@@ -48,18 +48,6 @@ let version = "21.1.8" in
       url = "gs://minimal-staging-archives/third-party-%{version}.src.tar.xz",
       sha256 = "7fe99424384aea529ffaeec9cc9dfb8b451fd1852c03fc109e426fe208a1f1a7"
     } | Source,
-    {
-      url = "gs://minimal-staging-archives/libcxx-%{version}.src.tar.xz",
-      sha256 = "6422a58a5c29b7f4fda224cfdc07842be8a208a61301bbba7a219116e3351809"
-    } | Source,
-    {
-      url = "gs://minimal-staging-archives/libcxxabi-%{version}.src.tar.xz",
-      sha256 = "709c9a63bde1e36a80d8675becc38073b85f0fa1b4111e34542b885c9e1239da"
-    } | Source,
-    {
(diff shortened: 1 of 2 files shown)


bryan-minimal:
Closing for now — landing this turned into a yak shave.

**What worked:** `LLVM_PARALLEL_LINK_JOBS=2` (link-phase OOM cap, distro-standard) and `build_args = { include version }` + version-templated build.sh (single-edit version bumps).

**What didn't:**

1. `LLVM_TARGETS_TO_BUILD` restriction → zig's cmake (`Findllvm.cmake`) requires LLVM to be built with all default targets enabled. Verification missed this — I checked "does any package use LLVM to cross-compile to backend X" (no) but not "does any package's BUILD require LLVM to be configured with backend X" (yes — zig).

2. `libcxx` + `libcxxabi` + `openmp` via `LLVM_ENABLE_RUNTIMES` → the standalone per-component LLVM tarballs don't reassemble into a layout that satisfies LLVM 21's `runtimes/CMakeLists.txt` path expectations. Three layouts tried (PROJECTS, RUNTIMES top-level, RUNTIMES under `<llvm-src>/runtimes/`), all failed differently.

**Right vehicle for a re-attempt:** monorepo tarball (`llvm-project-21.1.8.src.tar.xz`). It has the layout cmake expects out of the box AND collapses our 8 source tarballs to 1 (incidentally resolving [pkgmgr-rs#59](https://github.com/gominimal/pkgmgr-rs/issues/59) — multi-source updater gap). Will track as a separate issue.

The libcxx/libcxxabi/openmp source tarballs are already in `gs://minimal-staging-archives/` from this attempt; they're cheap and can stay or be cleaned up later.

PR #141, opened 2026-04-29
claude-code: add base as a runtime_dep instead of bash

twitchyliquid64:
I noticed that the `/login` failed to start the local listener unless we had `base`. Not 100% sure why, but for now I'm wiring `base` as a runtime_dep. Given `base` includes `bash`, im ripping that out.

## Chores
* Adjusted dependency configuration to streamline build and runtime package requirements.


bryan-minimal (approved):
LGTM

PR #155, opened 2026-05-06
Add chromium-bin and chromium-headless-shell-bin pkgs; agent-browser & mermaid-cli adopt them

agent-137:
## Summary

- Adds two prebuilt Chromium packages, sibling to each other and pinned via Playwright's revision system (rev 1217 / Chrome 147.0.7727.15):
  - `chromium-bin` — full headed browser → `/bin/chromium`
  - `chromium-headless-shell-bin` — stripped headless build used by Playwright/Puppeteer in default headless mode → `/bin/chromium-headless-shell`
- Per-arch sources: amd64 direct from Google's Chrome for Testing bucket (Playwright's CDN only redirects there for amd64 anyway, and that redirect is unreliable on newer revisions because MS's CDN doesn't host the amd64 zips), arm64 via Playwright's MS-hosted build.
- Both pkgs install into a shared `/usr/share/playwright-browsers/` directory using Playwright's registry layout, so installing both lets a consumer set `PLAYWRIGHT_BROWSERS_PATH` once and have `@playwright/test` discover both registry trees with no further config (provided the installed `playwright-core` revision matches).
- Stable wrappers at `/usr/bin/chromium` and `/usr/bin/chromium-headless-shell` so consumers don't need per-arch knowledge of the on-disk layout.
- `agent-browser` drops `npx playwright-core install chromium` and instead depends on `chromium-bin`; the daemon wrapper passes `--executable-path /usr/bin/chromium`.
- `mermaid-cli` drops its own headless-shell zip download and chromium system-lib runtime_deps; it depends on `chromium-headless-shell-bin` only and the `mmdc` wrapper sets `PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium-headless-shell`.
- READMEs in both pkg dirs document the two consumer routes for Playwright (`PLAYWRIGHT_BROWSERS_PATH` for drop-in when revisions match; `launchOptions.executablePath` when they don't).

## Naming: why `*-bin` and not `chromium` / `playwright-chromium`

`-bin` follows the Gentoo/AUR convention for a prebuilt binary distribution (vs. a source build). It signals what consumers actually get and leaves the bare `chromium` name free for a future source-built pkg. `playwright-chromium` would have been honest about provenance but read as "chromium for Playwright users" — which is wrong: mermaid-cli's adoption demonstrates Puppeteer-based consumers benefit too.

## Why two pkgs instead of one

The full headed Chromium and the headless-shell variant are two different upstream artifacts (Playwright models them as two separate browser installs). Shipping them together as one pkg meant every consumer paid the full ~290 MB even when only one variant was needed. mermaid-cli only needs the headless shell; agent-browser only needs the headed browser. Splitting matches Playwright's own model and lets consumers pull only what they use, while the shared `/usr/share/playwright-browsers/` registry dir keeps the zero-config `PLAYWRIGHT_BROWSERS_PATH` story intact for consumers that install both.

## Drop-in replacement for `npx playwright install chromium`

For Playwright users whose installed `playwright-core` revision matches the one shipped here, install both pkgs and:

```sh
export PLAYWRIGHT_BROWSERS_PATH=/usr/share/playwright-browsers
```

`@playwright/test` finds the pre-installed binaries at the layout it expects (`chromium-1217/<inner>/chrome` for headed, `chromium_headless_shell-1217/<inner>/{headless_shell,chrome-headless-shell}` for headless), skipping the download. `INSTALLATION_COMPLETE` markers in both browser dirs prevent re-fetch attempts.

When revisions don't match (e.g. consumer pins an older `playwright-core`), Playwright's auto-discovery walks a revision-suffixed path it won't find — and there's no `*_EXECUTABLE_PATH` env override for browsers in playwright-core. In that case the README documents the fallback: pass `launchOptions.executablePath: '/bin/chromium-headless-shell'` (or `/bin/chromium`) to bypass the registry lookup entirely.

Verified end-to-end on arm64 via a temp task (since removed): `chromium.launch()` resolves `executablePath` to the pkg's chrome, launches the headless-shell variant, returns the pinned Chrome version without any download attempt.

## Supply-chain wins versus runtime CDN install

- **Hash pinning is visible and auditable** — sha256s sit in `packages/chromium-{bin,headless-shell-bin}/build.ncl`, version-controlled and reviewable. `npx playwright install` does verify hashes, but the expected values are buried inside transitively-pinned npm deps.
- **Reproducibility tracks the pkg graph, not your `node_modules`** — two builds of the same pkg version always produce byte-identical chromium binaries.
- **No build-time network call** — Minimal fetches the zips into its content-addressed store once; subsequent users get cache hits. `npx playwright install` re-runs at every fresh `node_modules` install.
- **Per-arch dispatch is explicit in the spec** — the `match { arch = 'Amd64, .. } | { arch = 'Arm64, .. }` blocks make the cross-platform contract reviewable in one place.

## Why a prebuilt and not source

Chromium-from-source needs depot_tools, a hermetic clang toolchain, a ~30 GB git checkout, and 4–8 hours of build time — none of which is currently packaged. CLAUDE.md's "build from source" rule has an explicit carve-out for this case, called out in `chromium-bin/build.ncl`.

## Why the Playwright revision system instead of Chrome for Testing

Upstream agent-browser is built around Chrome for Testing (Google's official automation channel), and that would be the natural source. But as of 2026-05-06 Google does not publish a linux-arm64 CfT build — the [JSON manifest](https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions-with-downloads.json) has only `linux64`, `mac-arm64`, `mac-x64`, `win32`, `win64`. ARM64 Linux Chrome was [announced for Q2 2026](https://blog.chromium.org/2026/03/bringing-chrome-to-arm64-linux-devices.html) but the CfT publishing pipeline hasn't shipped arm64 zips yet ([upstream tracker](https://issues.chromium.org/issues/374811603)).

Playwright self-hosts a linux-arm64 chromium build, so its revision is the only pinnable chromium identifier today that works on both arches. amd64 still resolves to Google's CfT bucket — we just address it directly now (`storage.googleapis.com/chrome-for-testing-public/<version>/linux64/...`) instead of going through Playwright's CDN redirect, since that redirect is broken for newer revisions on amd64.

Notion follow-up tracks adding a separate `chrome-for-testing` pkg once Google ships arm64 CfT.

## Behavior change: mermaid-cli chromium revision bump

mermaid-cli previously pinned Playwright revision **1148** (Chrome ~141.x). After this PR it follows chromium-headless-shell-bin's revision **1217** (Chrome 147.0.7727.15). Headless rendering of typical mermaid diagrams should be identical, but font hinting / SVG layout edge cases between Chrome minor versions can exist — worth a render-diff smoke check before merging.

## Test plan

- [x] `min check --packages chromium-bin chromium-headless-shell-bin agent-browser mermaid-cli` clean
- [x] `min patched-pkg chromium-bin` and `chromium-headless-shell-bin` build and pass standalone smoketests (`--version`)
- [x] `min patched-pkg mermaid-cli` builds against `chromium-headless-shell-bin` and passes its install-check tests
- [x] `min patched-pkg agent-browser` builds against `chromium-bin` and `agent-browser --version` works
- [x] `@playwright/test` integration on arm64: `chromium.launch()` returns the pinned Chrome version without re-downloading
- [ ] mermaid-cli render smoke check after the rev 1148 → 1217 bump (e.g. `mmdc -i /tmp/example.mmd -o /tmp/out.svg` on a representative diagram, diff vs. known-good)
- [ ] amd64 build verification of full migration (only validated on arm64 host so far for downstream consumers)
- [ ] Webapp e2e migration to `chromium-bin`/`chromium-headless-shell-bin` (separate PR in webapp repo)

## Summary by CodeRabbit

* **Chores**
  * Refactored Chromium dependency management across packages to use a shared, pre-built distribution instead of individual installations, reducing build complexity and artifact size.
  * Simplified runtime dependencies by consolidating system library requirements into a single centralized package with multi-architecture support.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -711,3 +711,37 @@ export GONOSUMDB=*
 ```bash
 pip3 install --no-build-isolation --prefix=/usr --root=$OUTPUT_DIR .
 ```
+
+# You are in a Minimal execution environment
+
+You are running in an isolated sandbox/execution-environment specialized for software development. This environment has access to a few specialized tools to help you accomplish your goals.
+These tools are all subcommands of the `min` command.
+
+You will NOT be able to mutate the _host_ system with `apt-get`, `sudo pip install` etc, however you
+will be able to install additional commands using the `min add <packages>` command.
+
+## Adding packages: `min add <package name>`
+
+This command installs the package with the given name into your environment, letting you use the CLI tools or libraries it encapsulates.
+
+Without any flags, `min add` installs the given package for this session only. If you are sure that it will
+always be needed as a build-time (i.e. tools needed during compilation) or runtime dependency (i.e. dynamic
+libraries), you can add `--build` or `--runtime` to the invocation to depend on the package permanently.
+
+Help string for `min add`: min add [--session|--build|--runtime|--task <taskname>] <packages>
+
+Examples:
+
+ * The `curl` utility is needed: `min add curl`
+ * This project links against openssl: `min add --runtime openssl`
+ * This project has a build step requiring the protobuf compiler: `min add --build openssl`
+
+## Finding packages: `min search <search term>`
+
+Performs a text search across all available packages, printing packages that match or partially match the given term. Use this to find the exact names of packages you need.
+
+## Running tasks: `min run <task name>`
+
+This command runs the specified task (declared in this project's `minimal.toml`) in a separate Minimal execution environment, streaming back the output.
+
+Some typical tasks include `build` and `test`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -711,37 +711,3 @@ export GONOSUMDB=*
 ```bash
 pip3 install --no-build-isolation --prefix=/usr --root=$OUTPUT_DIR .
 ```
-
-# You are in a Minimal execution environment
-
-You are running in an isolated sandbox/execution-environment specialized for software development. This environment has access to a few specialized tools to help you accomplish your goals.
-These tools are all subcommands of the `min` command.
-
-You will NOT be able to mutate the _host_ system with `apt-get`, `sudo pip install` etc, however you
-will be able to install additional commands using the `min add <packages>` command.
-
-## Adding packages: `min add <package name>`
-
-This command installs the package with the given name into your environment, letting you use the CLI tools or libraries it encapsulates.
-
-Without any flags, `min add` installs the given package for this session only. If you are sure that it will
-always be needed as a build-time (i.e. tools needed during compilation) or runtime dependency (i.e. dynamic
-libraries), you can add `--build` or `--runtime` to the invocation to depend on the package permanently.
-
-Help string for `min add`: min add [--session|--build|--runtime|--task <taskname>] <packages>
-
-Examples:
-
- * The `curl` utility is needed: `min add curl`
- * This project links against openssl: `min add --runtime openssl`
- * This project has a build step requiring the protobuf compiler: `min add --build openssl`
-
-## Finding packages: `min search <search term>`
-
-Performs a text search across all available packages, printing packages that match or partially match the given term. Use this to find the exact names of packages you need.
-
-## Running tasks: `min run <task name>`
-
-This command runs the specified task (declared in this project's `minimal.toml`) in a separate Minimal execution environment, streaming back the output.
-
-Some typical tasks include `build` and `test`.
--- a/packages/chromium-bin/build.ncl
+++ b/packages/chromium-bin/build.ncl
@@ -1,4 +1,4 @@
-let { standaloneTest, Attrs, BuildSpec, Local, OutputBin, OutputData, Source, .. } = import "minimal.ncl" in
+let { standaloneTest, Attrs, BuildSpec, Local, OutputBin, OutputData, Source, Test, .. } = import "minimal.ncl" in
 let { target, .. } = import "config.ncl" in
 let base = import "../base/build.ncl" in
 let unzip = import "../unzip/build.ncl" in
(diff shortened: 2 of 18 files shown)


packages/chromium-bin/build.ncl
```diff
@@ -0,0 +1,184 @@
...
+
+  attrs =
+    {
+      upstream_version = browser_version,
+      source_provenance = {
+        category = 'GithubRepo,
+        owner = "microsoft",
+        repo = "playwright",
+      },
+      build_cost_multiple = 3,
```
twitchyliquid64: No need for a high build cost if this is just an unpack.

packages/chromium-bin/build.ncl
```diff
@@ -0,0 +1,184 @@
...
+
+  outputs = {
+    # /usr/bin/chromium               → full chromium browser (headed)
+    # /usr/bin/chromium-headless-shell → headless-shell variant (used by
+    #                                    Puppeteer/@playwright/test default headless)
+    bins = { glob = "usr/bin/chromium*" } | OutputBin,
+    data = { glob = "usr/share/chromium-bin/**", allow_executable = true } | OutputData,
+  },
+
+  attrs =
```
twitchyliquid64, line 141: Set the binary_from attribute for fetched binaries, see https://github.com/gominimal/minimal/blob/6f9be0a1d67f30ac131013c92dcc5a6af3c4bd0f/crates/stdlib/minimal-ncl/attr_classes.ncl#L63

packages/chromium-bin/build.ncl
```diff
@@ -0,0 +1,184 @@
...
+# doesn't, pass `executablePath: '/bin/chromium-headless-shell'` (or
+# '/bin/chromium' for headed) to bypass the registry lookup. See
+# packages/chromium-bin/README.md.
+let revision = "1217" in
+let browser_version = "147.0.7727.15" in
+{
+  name = "chromium-bin",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    # Full chromium browser (used in headed mode)
```
twitchyliquid64: Can we split the two versions of chrome into two different packages? otherwise this will be massive lol

mitodrummer: good call

CLAUDE.md
```diff
@@ -711,3 +711,37 @@ export GONOSUMDB=*
 ```bash
 pip3 install --no-build-isolation --prefix=/usr --root=$OUTPUT_DIR .
 ```
+
```
twitchyliquid64: I think this came across accidentally

We really need to use hooks XD

twitchyliquid64 (approved):

PR #208, opened 2026-05-29
feat(virtio-linux): EFI-bootable kernel + vmlinuz-efi output

norrietaylor:
Makes the `virtio-linux` kernel directly EFI-bootable, so it can boot under EFI firmware (e.g. OVMF, as used by `libkrun-efi`/`krunkit`) in addition to the existing direct-kernel-load VMMs.

**Changes:**
- Enable `CONFIG_EFI`, `CONFIG_EFI_STUB`, `CONFIG_ACPI`, `CONFIG_DMI`, `CONFIG_EFIVAR_FS` — adds the PE/COFF EFI-stub entry point and the ACPI/DMI support EFI firmware needs. Additive: the kernel still boots qemu/cloud-hypervisor/firecracker unchanged.
- New output `vmlinuz-efi`: the EFI-firmware-bootable image (uncompressed `Image` on arm64, since OVMF cannot unwrap `Image.gz`; the same `bzImage` on x86). The existing `vmlinuz` is unchanged.
- Bake a forced kernel cmdline (`CONFIG_CMDLINE_FORCE` + `console=hvc0 root=/dev/vda2 rootfstype=ext4 ro init=/sbin/stub-init`). OVMF launches the EFI-stub kernel with **empty** LoadOptions, so without a built-in cmdline the kernel has no `root=` and panics. This is a purpose-built microVM kernel for the Minimal VM image, whose disk convention is GPT { p1=ESP, p2=ext4 root } on `/dev/vda` — fair to bake. `init=/sbin/stub-init` is the current boot-verification entrypoint and becomes the in-VM agent (`/sbin/init`) later.

**Context:** consumed by the `alpine-virtio-linux` package, which assembles an EFI-bootable GPT disk (ESP + ext4 root) that krunkit boots.

**Validation:** validated end-to-end against krunkit 1.2.1 (libkrun-efi) on Apple Silicon — OVMF → kernel 6.12.43 → ext4 root → init. The `minimal build` CI check (external Linux build) is green, and `minimal check --packages virtio-linux` passes all 14 checkers locally (the artifact contains both `vmlinuz` and the EFI `vmlinuz-efi`).

### DIFF
--- a/packages/virtio-linux/build.ncl
+++ b/packages/virtio-linux/build.ncl
@@ -43,6 +43,7 @@ let version = "6.12.43" in
 
   outputs = {
     vmlinuz = { glob = "usr/share/virtio-linux/vmlinuz" } | OutputData,
+    vmlinuz_efi = { glob = "usr/share/virtio-linux/vmlinuz-efi" } | OutputData,
     config = { glob = "usr/share/virtio-linux/config" } | OutputData,
     system_map = { glob = "usr/share/virtio-linux/System.map" } | OutputData,
   },
--- a/packages/virtio-linux/build.sh
+++ b/packages/virtio-linux/build.sh
@@ -14,6 +14,10 @@ case "$(uname -m)" in
     # means the inner compression is gzip — no external wrap needed.
     KERNEL_TARGET=bzImage
     KERNEL_ARTIFACT=arch/x86/boot/bzImage
+    # bzImage already carries the EFI stub, so the EFI artifact is the same
+    # self-extracting image — no uncompressed variant needed on x86.
+    KERNEL_EFI_TARGET=bzImage
+    KERNEL_EFI_ARTIFACT=arch/x86/boot/bzImage
     HAS_KVM_GUEST_FRAGMENT=1
     ;;
   aarch64)
@@ -21,6 +25,10 @@ case "$(uname -m)" in
     # that qemu/cloud-hypervisor/firecracker all accept.
     KERNEL_TARGET=Image.gz
     KERNEL_ARTIFACT=arch/arm64/boot/Image.gz
+    # EFI firmware (e.g. OVMF) loads the kernel as a PE/COFF application and
+    # cannot unwrap the gzip, so the EFI artifact is the uncompressed Image.
+    KERNEL_EFI_TARGET=Image
+    KERNEL_EFI_ARTIFACT=arch/arm64/boot/Image
     HAS_KVM_GUEST_FRAGMENT=0
     ;;
   *)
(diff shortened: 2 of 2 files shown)


packages/virtio-linux/build.sh
```diff
@@ -135,16 +143,51 @@ $CFG --enable KVM_AMD
...
+# LoadOptions, so without a built-in cmdline the kernel has no root= and panics
+# ("Unable to mount root fs"). CMDLINE_FORCE overrides the empty bootloader
+# cmdline rather than the default *_FROM_BOOTLOADER (which would defer to the
+# empty one).
+#
+# init=/sbin/stub-init is the current boot-verification entrypoint; it becomes
+# the in-VM agent (/sbin/init) in a later iteration.
+$CFG --enable CMDLINE_BOOL
+$CFG --set-str CMDLINE "console=hvc0 root=/dev/vda2 rootfstype=ext4 ro init=/sbin/stub-init"
+$CFG --enable CMDLINE_FORCE
```
norrietaylor, line 171: Baking the cmdline here ties this kernel to the VM-image disk convention (`root=/dev/vda2`, `init=/sbin/stub-init`). That's intentional for a purpose-built microVM kernel and needs zero extra packages.

The alternative, if this kernel ever needs to stay generic (multiple image variants with different cmdlines, an initrd, or secure/measured boot): drop `CMDLINE*` here and supply the cmdline at the image-assembly layer via a **UKI** — `ukify`/systemd-stub embeds a `.cmdline` section into the EFI binary, set per-image without recompiling the kernel. Costs a systemd-stub package (+ deps), so deferred until one of those needs it. `init=/sbin/stub-init` also becomes the in-VM agent (`/sbin/init`) later.

twitchyliquid64: You can typically also set the kernel command line when you load the kernel, so i guess most of the time this is mostly a default

packages/virtio-linux/build.ncl
```diff
@@ -43,6 +43,7 @@ let version = "6.12.43" in
 
   outputs = {
     vmlinuz = { glob = "usr/share/virtio-linux/vmlinuz" } | OutputData,
+    vmlinuz_efi = { glob = "usr/share/virtio-linux/vmlinuz-efi" } | OutputData,
```
twitchyliquid64, line 46: Do we want to just ship vmlinuz_efi and skip the non-efi one, if we need efi for minvmd?

norrietaylor:
Closing as not planned - I wanted to use krunkit with this. I have since changed my mind. In the future if someone wants to use krunkit perhaps that is a different variant.

PR #229, opened 2026-06-07
add microvm-rootfs and virtio-kernel-raw packages

norrietaylor:
Two packages for running a libkrun-based Linux microVM.

## virtio-kernel-raw
Gunzips the `virtio-linux` `Image.gz` to an uncompressed aarch64 `Image`, which libkrun can load with `KRUN_KERNEL_FORMAT_RAW` — skipping its in-VMM gzip decompress for a faster microVM boot. build_deps: virtio-linux, gzip, bash, coreutils.

## microvm-rootfs
A read-only ext4 guest rootfs: snapshots the closure of socat + bash + coreutils, adds a small vsock bring-up init (`/sbin/microvm-init`: writes a `READY` marker on vsock 7350, then serves an echo on vsock 2222) and an `/etc/microvm/manifest` contract, prunes build-only bulk, and packs an ext4 image (`mke2fs -O ^has_journal`) to load as a virtio-blk block device. build_deps: socat, bash, coreutils, e2fsprogs.

## Validation
`minimal check microvm-rootfs virtio-kernel-raw` — all checkers pass (parse, import line, imports, fmt, naming, executable scripts, disallowed-patterns).

## Summary by CodeRabbit

* **New Features**
  * Produce raw virtio kernel images with automatic handling of compressed or uncompressed kernel artifacts.
  * Produce a read-only ext4 microVM root filesystem image suitable for microVM guests.
  * MicroVM rootfs includes a tiny init for readiness signaling, trimmed runtime contents, and automated sizing for packaging.


packages/microvm-rootfs/build.ncl
```diff
@@ -0,0 +1,27 @@
...
+  name = "microvm-rootfs",
+
+  # A read-only ext4 guest rootfs for a libkrun microVM. build.sh snapshots the
+  # closure of these build_deps (socat + bash + coreutils + their libs;
+  # e2fsprogs provides mke2fs at build time), drops in a small bring-up init,
+  # prunes build-only bulk, and packs an ext4 image to load as a virtio-blk
+  # block device.
+  build_deps = [
+    { file = "build.sh" } | Local,
+    socat,
```
twitchyliquid64: What is socat used for in the build script?

packages/microvm-rootfs/build.ncl
```diff
@@ -0,0 +1,27 @@
...
+    bash,
```
twitchyliquid64: Might make sense to depend on base instead for build_deps

packages/microvm-rootfs/build.ncl
```diff
@@ -0,0 +1,27 @@
...
+    coreutils,
+    e2fsprogs,
+  ],
+
```
twitchyliquid64, line 22: I think it would be more pure to depend on the things you need in the rootfs image as runtime_deps, as that’s kinda the vibe of runtime deps. Sounds like you probably want base, socat, core utils as runtime deps and everything else as build deps, but note in practice the union of both will show up in your rootfs.

twitchyliquid64 (approved):
I’ll approve it tho so your g2g when you’re happy with it

packages/microvm-rootfs/build.ncl
norrietaylor: it is placeholder will replace it later.

norrietaylor:
Addressed the feedback (commits 7133c1e, 3cd24e6):

**@twitchyliquid64 (dep hygiene)** — `microvm-rootfs` now declares `runtime_deps = [base, socat]` and `build_deps = [build.sh, e2fsprogs]` (e2fsprogs is build-only: it provides `mke2fs` to pack the image). `base` covers bash/coreutils/gzip.

**CodeRabbit:**
- *Build-only tools in the image* — the staged tree drops the e2fsprogs files (`usr/sbin`, the few `usr/bin` tools, `libext2fs/libe2p/libss`) before `mke2fs` runs (which executes from the sandbox PATH, not `$STAGE`). Verified against the built image: no `mke2fs`/`e2fsck` in `/usr/sbin`, e2fs libs gone, `socat`/`microvm-init`/`bin/sh` intact.
- *Init never failing on a missed READY* — added `[ "$i" -lt 50 ] || { echo …; exit 1; }` before the listener exec.
- *`|| true` swallowing the whole prune block* — scoped to the `find` only; `cd`/`rm -rf` failures now fail the build.
- *virtio-kernel-raw missing version plumbing* — added `version`, `attrs.upstream_version`, and `build_args { include version }`.

Both packages `minimal package --rebuild` clean.

PR #238, opened 2026-06-11
glibc 2.42→2.43 + C23-const FTBFS fixes (make stack + gcc)

bryan-minimal:
## What

glibc **2.42 → 2.43**, plus every FTBFS fix the full-fleet rebuild surfaced. Reshaped to **glibc-only** (binutils 2.46.1 already landed separately on main, so it's dropped here) and rebased onto current main (picks up the harnesses → `stacks/` migration). **Builds clean on amd64 + arm64.**

## Why

glibc 2.43 clears the glibc CVE backlog (CVE-2025-15281 + CVE-2026-5450 CRIT, etc.). Because glibc is libc, the bump rebuilds the entire fleet — which is what surfaced the FTBFS wave below.

## The C23-const FTBFS wave

glibc 2.43 implements ISO C23 const-preserving string macros — `strchr`/`strrchr`/`strstr`/`memchr`/`bsearch`/etc. now return `const char *` for a `const char *` argument. Pre-C23 code assigning the result to a plain `char *` trips **`-Werror=discarded-qualifiers`** and FTBFS. This is an ecosystem-wide wave, and our fleet hits it in two distinct shapes:

**1. The make-stack lever** — `stacks/make/stack.ncl` appends `-Wno-error=discarded-qualifiers` to the shared make-stack CFLAGS; it lands after a package's own `-Werror` (later flag wins), downgrading that one warning.

> ⚠️ **Honest scope note:** this lever only reaches packages that actually use the make stack. Most C packages here have a **hand-written `build.sh` that assigns its own `CFLAGS`** (an assignment, not an append), so they **bypass** the make-stack lever entirely. There's no in-repo universal interception point (the global `-fno-plt` etc. come from the build sandbox, not this repo). So the lever is correct + durable for make-stack packages, but the packages below needed individual fixes — surfaced one-at-a-time by the buildbot as it rebuilt the fleet.

**2. Per-package `build.sh` fixes** — for the hand-written-`build.sh` packages that also self-enable `-Werror` (their `-Werror` lives in upstream configure, invisible to a grep), the same `-Wno-error=discarded-qualifiers` goes into each package's own CFLAGS. It lands after their `-Werror` on the compile line (or after `$(AM_CFLAGS)` under automake), so the later flag wins.

**3. or-tools — a different class entirely** (not `-Werror`, a hard compile error). or-tools FetchContent's SCIP v10.0.0, which bundles **tinycthread** (a C11 `<threads.h>` shim doing `#define once_flag pthread_once_t`). glibc 2.43 surfaces its native `once_flag`/`call_once` into SCIP's translation unit, so that macro corrupts glibc's `typedef __once_flag once_flag;` into a `pthread_once_t` redefinition → "conflicting types". `-Wno-error` can't fix a hard error. or-tools 9.15 is already the latest release and hard-`FORCE`s SCIP `TPI=tny` in its bundled cmake (so a `-DTPI=...` is overridden), so we sed that to **`TPI=omp`** before cmake — SCIP uses its OpenMP task interface instead of tinycthread, staying parallel + THREADSAFE (libgomp is present from our gcc build). A `grep` guard fails the build loudly if upstream renames the line.

## Changes

| File | Change | Kind |
|------|--------|------|
| `packages/glibc/build.ncl` | 2.42 → 2.43 (+ sha256) | the bump |
| `stacks/make/stack.ncl` | fleet `-Wno-error=discarded-qualifiers` | durable (make-stack pkgs) |
| `packages/gcc/build.sh` | same downgrade via `CFLAGS_FOR_TARGET` (libgomp) | **interim** (retired by gcc-16) |
| `packages/libxcrypt/build.sh` | `-Wno-error=discarded-qualifiers` in CFLAGS (crypt-{gost,sm3}-yescrypt.c) | per-package |
| `packages/elfutils/build.sh` | `-Wno-error=discarded-qualifiers` in CFLAGS (libcpu/riscv_disasm.c) | per-package |
| `packages/strace/build.sh` | `-Wno-error=discarded-qualifiers` in CFLAGS (src/ioctl.c) | per-package |
| `packages/or-tools/build.sh` | SCIP `TPI` tny → omp (drop tinycthread) | tinycthread collision |

## For the reviewer (parked / notable)

- **gcc's fix is interim.** A gcc-16 bump retires `packages/gcc/build.sh` (gcc-16 patched its own libgomp upstream) but does **not** retire the fleet/per-package fixes — every other C program still trips the same break under any gcc against glibc 2.43, because the language semantics are unchanged.
- **The real toolchain-policy call** is `-Wno-error=discarded-qualifiers` (downgrade the one warning, used here) vs **`-std=gnu17`** (restores C17 prototypes so `strchr` returns `char *` — no warning at all, but a broader language-standard change). This PR takes the surgical `-Wno-error` route; the `-std=gnu17` switch can replace all of these levers later if the team prefers it.
- **or-tools `TPI=omp` tradeoff:** SCIP keeps parallelism via OpenMP instead of tinycthread. If OpenMP were ever unavailable it fails fast at *configure* (not after the long compile), and the fallback is `TPI=none` (single-threaded SCIP).
- **elfutils gotcha** (for anyone touching it later): its configure has **no `--disable-werror`** in this version — the CFLAGS flag is the reliable knob.

## Status

**Draft** — builds clean on both arches, but it's a toolchain-policy change (touches the C-build warning posture fleet-wide), so leaving it for explicit sign-off on the `-Wno-error` vs `-std=gnu17` direction before merge.


bryan-minimal:
### Added: gcc glibc-2.43 build fix (commit 4cf505f)

The first buildbot run failed building **gcc-15.2.0** against the new **glibc 2.43** headers — not glibc/binutils themselves. glibc 2.43 implements [ISO C23 const-preserving string macros](https://lists.gnu.org/archive/html/info-gnu/2026-01/msg00005.html): `strchr`/`strrchr`/`strstr`/`memchr`/… now return `const char *` for a `const char *` input. gcc-15.2.0 predates this, so `libgomp/affinity-fmt.c` (`char *q = strchr(p+1,'}')`) discards `const` → `-Werror=discarded-qualifiers` → FTBFS.

This is the **same C23 wave hitting the whole ecosystem right now** (openssl, grub, libevent, criu, …) — a known, documented event, not anything specific to us.

**Fix here (interim, surgical):** `-Wno-error=discarded-qualifiers` via `CFLAGS_FOR_TARGET` (the [documented knob](https://gcc.gnu.org/install/build.html) for target libs like libgomp) for gcc's *own* build only. It downgrades exactly that one warning — everything else stays `-Werror` — and the shipped compiler is byte-identical. The pointer is only used for `q - p`, never written, so it's source-pedantry, not a runtime bug.

**Clean follow-up:** bump gcc → **16.1.0**, which fixes this upstream and builds against glibc 2.43 with no workaround ([LFS pairs them directly](https://www.linuxfromscratch.org/lfs/view/development/chapter08/gcc.html)). That retires the band-aid; tracked separately as it's a major-version migration.

Refs are also inline in `packages/gcc/build.sh`.

bryan-minimal:
## 🚧 Moved to draft — needs a team decision on the glibc-2.43 C23 wave

Second buildbot run got past gcc but then **libxcrypt** failed with the *identical* pattern:
```
lib/crypt-gost-yescrypt.c:134: error: initialization discards 'const' qualifier ... [-Werror=discarded-qualifiers]
  char *hptr = strchr ((const char *) intbuf->retval + 3, '$');
```
So this is confirmed **fleet-wide**, not a gcc one-off (see the [previous comment](#) for root cause: glibc 2.43's ISO C23 const-preserving string macros). Patching package-by-package is whack-a-mole, and **gcc 16.1.0 would *not* help** — it only fixes gcc's own source, not libxcrypt/openssl/grub/etc.

### The systematic fix: one line in the shared make harness

The flags that land *after* each package's `-Werror` come from `harnesses/make/harness.ncl`:
```nickel
# harnesses/make/harness.ncl:9
CFLAGS = "-O2 -pipe -gno-record-gcc-switches",
```
Because these are appended last on every compile (confirmed in libxcrypt's command line: `… -Werror -fno-plt -march=… -O2 -pipe -gno-record-gcc-switches …`), adding the override here downgrades the one warning **fleet-wide for every make/autotools package, in a single line**:
```diff
-    CFLAGS = "-O2 -pipe -gno-record-gcc-switches",
+    CFLAGS = "-O2 -pipe -gno-record-gcc-switches -Wno-error=discarded-qualifiers",
```
gcc has a custom `build.sh`, so it's handled separately (the commit already in this PR). A handful of CMake/meson packages have their own harnesses — if any of *those* break, the same one-line override applies to `harnesses/cmake` / `harnesses/meson`.

### Why this is safe (but is a band-aid)
- Downgrades **exactly one** warning (`discarded-qualifiers`) error→warning; every other `-Werror` stays.
- The flagged code is **runtime-safe** — it's the old C17 `char *q = strchr(const…)` pattern where the pointer isn't written; byte-identical output.
- It does **not** change the language standard or any ABI.
- **Tech debt:** it *masks* the not-yet-const-correct code rather than fixing it. We unwind it when we adopt C23 properly.

### Decision for the team
| Option | Effect | Cost |
|---|---|---|
| **Global `-Wno-error=discarded-qualifiers` in harness** (proposed) | Whole wave gone, glibc 2.43 + CVEs land now | Band-aid / tech debt |
| **`-std=gnu17` in harness** | Broader — turns off *all* C23, not just this warning | Bigger hammer; also masks future C23 issues |
| **Per-package `const` patches** (what distros/upstreams do) | "Correct" C23-forward | Whack-a-mole across N packages |
| **Hold glibc 2.43** | Safe, no churn | glibc CVEs stay unpatched |
| **Bump gcc → 16.1.0** | Fixes *gcc only* — does NOT address libxcrypt/others | Major-version migration, doesn't solve the wave |

**My recommendation:** the global `-Wno-error=discarded-qualifiers` harness override is the pragmatic way to land glibc 2.43 (and clear the CVEs) **now** without whack-a-mole, then do the const-correct per-package fixes (or a gcc-16 bump) as deliberate follow-up. But this touches *every* C build, so it's a deliberate toolchain-policy call — hence draft + this writeup for discussion.

Refs: [glibc 2.43 C23 const macros](https://lists.gnu.org/archive/html/info-gnu/2026-01/msg00005.html) · [GCC CFLAGS_FOR_TARGET docs](https://gcc.gnu.org/install/build.html) · [LFS GCC-16.1.0](https://www.linuxfromscratch.org/lfs/view/development/chapter08/gcc.html)

msample (approved):
LGTM

twitchyliquid64 (approved):

PR #240, opened 2026-06-12
Update binutils 2.46.1 + coreutils 9.11 (base-soup bundle)

bryan-minimal:
Bundles two base-soup toolchain updates so the cascading rebuild lands as one unit:
- **binutils** 2.46.0 → 2.46.1
- **coreutils** 9.10 → 9.11

**glibc** has an update too but is **deferred** to a separate change (per request).

Both mirrored + **sha-verified against upstream ftp.gnu.org** (.tar.xz):
| pkg | sha256 | matches upstream |
|---|---|---|
| binutils-2.46.1.tar.xz | `e127a709…` | ✅ |
| coreutils-9.11.tar.xz | `394024ed…` | ✅ |

⚠️ Heads-up for the reviewer: coreutils' tarball was **manually verified/corrected** — pkgmgr's GNU-FTP parser grabbed the `.tar.gz` when `.tar.xz` was also available, so I re-mirrored + sha-checked the `.xz` by hand. A parser-preference fix (prefer `.tar.xz`) is a separate pkgmgr follow-up.

## Summary by CodeRabbit

* **Chores**
  * Updated binutils to version 2.46.1
  * Updated coreutils to version 9.11

### DIFF
--- a/packages/binutils/build.ncl
+++ b/packages/binutils/build.ncl
@@ -16,13 +16,13 @@ let glibc = import "../glibc/build.ncl" in
 let linux_headers = import "../linux_headers/build.ncl" in
 let zlib = import "../zlib/build.ncl" in
 
-let version = "2.46.0" in
+let version = "2.46.1" in
 {
   name = "binutils",
   build_deps = [
     {
       url = "gs://minimal-staging-archives/binutils-%{version}.tar.xz",
-      sha256 = "d75a94f4d73e7a4086f7513e67e439e8fcdcbb726ffe63f4661744e6256b2cf2",
+      sha256 = "e127a709cba24c76de8936cb7083dd768f28cd37eb010492e2f19b71eb1294e4",
       extract = true,
     } | Source,
     { file = "build.sh" } | Local,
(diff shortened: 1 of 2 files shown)


bryan-minimal:
Superseded by #242 — closing.

The binutils 2.46.0→2.46.1 + coreutils 9.10→9.11 bumps from this PR are carried verbatim in #242 (commit `34cfe32`, the base of that branch). #242 then stacks the fixes the cascade rebuild turned out to need:

- **libvips 8.18.2 → 8.18.3** — sharp 0.35.1 (pulled unpinned by `next/build.sh`) hard-requires libvips ≥ 8.18.3, released 2026-06-09.
- **next: glib build/runtime dep + surface glib include paths to the sharp native build** — libvips's public `vips/vips8` header pulls in `<glib-object.h>`, which `pkg-config vips-cpp` doesn't expose (glib is a `Requires.private`).

Splitting these out would just re-break the same cascade, so per the bundle-the-cascade convention they ride together in #242. Landing the whole base-soup bundle (binutils + coreutils + libvips + next) as one unit there.

PR #276, opened 2026-06-19
python: make build reproducible (libffi pin + drop PGO)

bryan-minimal:
## What

Makes the `python` package build byte-reproducibly. Two independent fixes:

1. **Pin libffi's pkg-config vars** (`LIBFFI_CFLAGS=-I/usr/include LIBFFI_LIBS=-lffi`) so `configure` resolves `MODULE__CTYPES_LDFLAGS` deterministically as `-lffi`, instead of letting `PKG_CHECK_MODULES` non-deterministically pick `-L/usr/lib/../lib64` vs `-L/usr/lib/../lib` — which leaked into `_sysconfigdata*.py/.json/Makefile` (6 files). Verified eliminated in build-twice.

2. **Drop `--enable-optimizations` (PGO)** and add `-fno-semantic-interposition` to recover the perf.

## Why drop PGO

`--enable-optimizations` runs a gcc profile-guided-optimization training pass whose `.gcda` counters are **not reproducible**. Measured directly: build the instrumented interpreter once, run `-m test --pgo` twice from a clean `.gcda` slate, and **216 of 334 `.gcda` files differ (64%)** — across the core interpreter, parser, and compiler. Some even differ in *size* (different functions execute run-to-run), and this is with the hash seed already pinned by the sandbox. No gcc flag fixes the training *data*, and the profile workload can't be made deterministic without checking in a stored profile (a maintenance/footgun cost we don't want).

Dropping PGO makes `.text` byte-identical across builds. `-fno-semantic-interposition` (added to `CFLAGS` + `LDFLAGS`) recovers the main *deterministic* perf win `--enable-optimizations` bundled for `--enable-shared` — it drops PLT indirection on internal `libpython` calls — so the net perf cost over our `-O3 -march` baseline is modest (~3–7%).

This matches the reproducible-build choice already made by **Nixpkgs, openSUSE, and Yocto**, all of which drop CPython PGO for determinism.

## Verification

Build-twice + `repro-check diff`: functional sections (`.text/.rodata/.data`, incl. `libpython`) are **byte-identical**. The only residual diff observed was a DWARF `.debug_str` assembler-version string (`GNU AS 2.46.0` vs `2.46.1`) — environmental toolchain drift from binutils not being pinned across the two build windows, **not** a python source issue. Under a pinned toolchain in clean CI, python is byte-reproducible.

`minimal check --packages python` passes (incl. `build script disallowed-patterns`).

> Note on LTO: `--with-lto=full` is also non-deterministic here (WHOPR partitioning perturbs `libpython` layout build-to-build); not added. `-flto=1` (serial) would be deterministic if LTO perf is wanted later.

## Summary by CodeRabbit

* **Chores**
  * Updated Python build configuration to improve build determinism through environment variable pinning and compiler flag adjustments.

### DIFF
--- a/packages/python/build.sh
+++ b/packages/python/build.sh
@@ -9,19 +9,30 @@ case $(uname -m) in
   aarch64) MARCH="-march=armv8-a" ;;
   *)       MARCH="" ;;
 esac
-export CFLAGS="$MARCH -O3 -pipe -gno-record-gcc-switches -ffile-prefix-map=$(pwd)=/builddir"
-export LDFLAGS="-Wl,--build-id=none"
+# -fno-semantic-interposition: drops PLT indirection on internal libpython calls
+# — a deterministic perf win for --enable-shared that --enable-optimizations used
+# to add implicitly. Re-added by hand since we drop --enable-optimizations below.
+export CFLAGS="$MARCH -O3 -pipe -gno-record-gcc-switches -ffile-prefix-map=$(pwd)=/builddir -fno-semantic-interposition"
+export LDFLAGS="-Wl,--build-id=none -fno-semantic-interposition"
 export CXXFLAGS="${CFLAGS}"
 
-# Reproducibility: pin libffi's pkg-config vars so configure resolves
-# MODULE__CTYPES_LDFLAGS deterministically ('-lffi') rather than letting
-# PKG_CHECK_MODULES non-deterministically pick `-L/usr/lib/../lib64` vs
-# `-L/usr/lib/../lib` (which leaks into _sysconfigdata*.py/.json/Makefile).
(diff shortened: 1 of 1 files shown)


bryan-minimal:
Marking this **draft** pending the full python reproducibility picture. This PR fixes only the libffi `lib64/lib` flip (one of three determinism issues, confirmed eliminated). Full byte-reproducible python additionally needs: (a) a decision on `--enable-optimizations`/PGO (training data varies run-to-run — a perf tradeoff), and (b) a pinned build toolchain (`.comment` records the binutils version, which drifted 2.46.0→2.46.1 across verify builds). Will un-draft once those are settled and the whole package is verified byte-identical.

mitodrummer (approved):

PR #277, opened 2026-06-19
ghc: deterministic link order + package.cache (reproducible Haskell)

bryan-minimal:
## What

Makes `ghc` (GHC 9.10.3) reproducible — and, because the fix is in the compiler/linker, it makes **every downstream Haskell package** (stack, haskell-language-server, …) reproducible too.

## Diagnosis

GHC's *functional* output is already reproducible — `.text`/`.rodata`/`.data` are byte-identical across builds (even the 71 MB `libHSghc.so`). The entire residual is **metadata ordering**, from GHC iterating package/UnitId collections in hash-set order (`nonDetEltsUFM`), in two places:

1. **Link-arg / `DT_NEEDED` / `.dynstr` ordering** in every linked binary (also what makes `stack`'s identically-sized functions land at different addresses).
2. **`ghc-pkg recache`** → `package.cache` serialized in unsorted `readdir` order.

`-fobject-determinism` is *not* the lever here (codegen is already deterministic).

## Fix (two source patches, applied before the Hadrian build)

1. **Link order** — exact backport of upstream **GHC #26838 / MR !15453** (a single `sort $` on `preload1` in `compiler/GHC/Unit/State.hs`). Merged upstream for 10.0.1; **not** in 9.10.x; **Debian ships this exact patch for 9.10.3** (Debian #1125305). It restores the sorted-by-UnitId order GHC ≤ 9.6 had — behavior-restoring, low-risk.
2. **`package.cache`** — sort the `.conf` list before it's read/serialized in `utils/ghc-pkg/Main.hs` (`sort` already imported). No upstream fix exists; trivial and additive. The build's existing post-install `ghc-pkg recache` then emits a byte-identical cache.

Each patch is **guarded by a `grep`** that fails the build loudly (in seconds) if a future GHC moves the target, rather than silently regressing.

## Status — DRAFT pending verification

The patches are high-confidence (Debian-proven for this exact version, applied behind fail-fast guards) but **not yet locally build-twice-verified** — a from-scratch GHC build is multi-hour (`build_cost_multiple=6`), ×2 for a repro check. Marking draft until that verify (or CI / a beefy builder) confirms byte-identical output + `ghc --version`.

When minimal moves to GHC ≥ 10.0.1, drop patch (1) (it lands upstream); keep (2) (still unfixed upstream).

## Summary by CodeRabbit

* **Chores**
  * Improved build reproducibility by applying patches that ensure deterministic generation of GHC build outputs, with enhanced error checking to validate patch application.

### DIFF
--- a/packages/ghc/build.sh
+++ b/packages/ghc/build.sh
@@ -31,6 +31,10 @@ sed -i 's/let preload1 = nonDetKeysUniqMap (filterUniqMap (isJust . uv_explicit)
 ghc_pkg_main=utils/ghc-pkg/Main.hs
 grep -q 'confs = map (path </>) $ filter (".conf" `isSuffixOf`) fs' "$ghc_pkg_main" \
   || { echo "ERROR: ghc-pkg package.cache patch target not found in $ghc_pkg_main — GHC source changed." >&2; exit 1; }
+# The `sort $` below needs `sort` in scope; 9.10.3's Main.hs imports it (line 78),
+# but guard it so a future GHC import reorg fails here, not deep in the build.
+grep -qE 'import Data.List \(.*\bsort\b' "$ghc_pkg_main" \
+  || { echo "ERROR: 'sort' not imported in $ghc_pkg_main — patch (2) requires it (add a sort import)." >&2; exit 1; }
 sed -i 's#confs = map (path </>) $ filter (".conf" `isSuffixOf`) fs#confs = map (path </>) $ sort $ filter (".conf" `isSuffixOf`) fs#' "$ghc_pkg_main"
 
 # Create stubs for tools ./configure / bootstrap.py checks for but aren't strictly needed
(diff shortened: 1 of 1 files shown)


bryan-minimal:
Good catch, thanks — agreed, the patch-(2) guard checks the target line but not that `sort` is importable. For **GHC 9.10.3 specifically it's safe**: `utils/ghc-pkg/Main.hs:78` already has `import Data.List ( group, sort, sortBy, nub, partition, find … )`, so `sort` is in scope and the patch compiles (verified against the cached 9.10.3 source). So the current build-twice (running now) is fine.

I'll add the fail-fast import guard to harden against a future GHC reorg — e.g. `grep -qE 'import Data.List \(.*\bsort\b' "$ghc_pkg_main"` before the sed (or mirror patch (1) and `sed` the import in if absent). I'm holding the edit until the in-flight overnight build-twice finishes, so the two builds use an identical build.sh; will fold it in when I un-draft.

bryan-minimal:
✅ **Verified reproducible.** Overnight build-twice (aarch64, 8 CPU/28G) came back **byte-identical: 6602/6602 files** (`repro-check diff`). Both patches confirmed working. Un-drafted, and folded in the `sort`-import guard from the review above (output-neutral, so the verified result stands). This also unblocks the rest of the Haskell tail — stack and HLS build on this compiler.

mitodrummer (approved):

PR #279, opened 2026-06-20
ci: add version-age report-only check (inbox#21/#20)

bryan-minimal:
## What
A **report-only, non-blocking** version-age check for branch promotion (gominimal/inbox#21, criterion in #20).

- `.github/workflows/version-age.yml` — runs on `pull_request`, `contents: read`, no secrets.
- `.github/scripts/version_age_report.py` — for each changed `packages/*/build.ncl`, resolves the upstream release date (`attrs.released_at` → GitHub releases/tags → `ftp.gnu.org` `Last-Modified` → unknown), computes age, posts a job-summary table flagging anything <7 days old.

## Why it's safe (two-way door)
Purely additive new files; modifies nothing. Read-only token, no secrets, **never fails the job** (the script always exits 0; the workflow downgrades errors to warnings). Doesn't touch the cache, builds, branches, the merge queue, or any user pin. Revert = delete the two files.

> **Keep it OFF branch-protection required checks** — that exclusion is the only thing that guarantees non-blocking (a workflow can't self-enforce it).

## Honest scope
The real value isn't the pass/fail icons — it's the live **census of packages with no derivable date** (~82 with no `source_provenance`: toolchains, X11 libs, pinned binaries) that would need an `attrs.released_at` convention. GNU dates *are* derivable via `ftp.gnu.org`, so the underivable set is smaller than first thought. It's a *soak* gate (don't promote something released hours ago), not a staleness gate.

## Deferred (not in this PR)
The `attrs.released_at` schema (a separate `minimal` PR) + the example + `minimum_version` bump are gated on a one-shot census sizing the ~82-package backfill (to be posted to #20). The script handles `released_at`'s absence gracefully, so this PR stands alone.

Refs gominimal/inbox#21, gominimal/inbox#20.

## Summary by CodeRabbit

* **Chores**
  * Added an automated, report-only version-age check in pull requests that generates a Markdown table in the job summary for changed packages, including release dates, age verdicts, and source details.
  * Implemented hardened date resolution (multiple resolution strategies and safer network/provenance handling) and non-blocking behavior when dates can’t be determined.

### DIFF
--- a/.github/scripts/version_age_report.py
+++ b/.github/scripts/version_age_report.py
@@ -9,19 +9,22 @@ programming error, which the workflow downgrades to a warning.
 
 Date resolution precedence, per package:
   1. attrs.released_at            -- explicit maintainer declaration (override)
-  2. GithubRepo  -> GitHub API    -- releases-by-tag, else tag's commit date
-  3. GnuProject  -> ftp.gnu.org   -- Last-Modified of the release tarball
-  4. any source  -> source-URL    -- HEAD the package's own source for its
+  2. self-dated version           -- trailing YYYYMMDD in the version (ncurses &
+                                     other dated snapshots); pure, no network
+  3. GithubRepo  -> GitHub API    -- releases-by-tag, else tag's commit date
+  4. GnuProject  -> ftp.gnu.org   -- Last-Modified of the tarball (flat / nested
+                                     per-version dir / aliased project dir)
+  5. any source  -> source-URL    -- HEAD the package's own https source for its
                                      Last-Modified ("date on the file"); catches
-                                     no-provenance packages + tiers 2/3 misses
-  5. otherwise   -> UNKNOWN       -- "needs attrs.released_at" (non-http source)
+                                     no-provenance packages + tier 3/4 misses
+  6. otherwise   -> UNKNOWN       -- "needs attrs.released_at"
 
-Tier 4 shrinks the UNKNOWN set: most no-provenance packages (toolchains, X11
-libs, pinned binaries) still have an http(s) source tarball whose Last-Modified
-is a usable availability date. The genuine residue is sources with no http(s)
-date (gs:// mirrors, git), which still need attrs.released_at. (Note: Last-
-Modified is the artifact's availability/upload time, not strictly the upstream
-release date -- close enough for a soak gate, and dwell backstops it.)
+A full-catalog census (2026-06-20) put tier coverage at 280/373: most sources
+are a `gs://minimal-staging-archives/` mirror (no HTTP Last-Modified oracle), so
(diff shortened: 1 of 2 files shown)


bryan-minimal:
## Full-catalog run + empirical matcher — what landed, and deferred follow-ups

Ran the resolver over the whole catalog (373 pkgs); the run drove the two most recent commits. Coverage is now **279/373 auto-dated**. An empirical probe of *every* dateless-but-has-provenance package (real GitHub/GitLab tags APIs + GNU ftp listings) fed the new `candidate_tags` / `gnu_date` / self-dated tiers, reclaiming **17 of 20**.

**Verified reclaims** (exact upstream ref → date, all confirmed live):
openssh `V_10_3_P1`, postgres `REL_18_4`, tcl `core-8-6-16`, llvm `llvmorg-21.1.8`, sqlite `version-3.50.4`, codex `rust-v0.130.0`, gopls `gopls/v0.21.1`, cabal `cabal-install-v3.12.1.0`, lcms2 `lcms2.17`, bun `bun-v1.3.14`, varlock `varlock@0.2.3`, libfuse `fuse-3.18.2`, libxkbcommon `xkbcommon-1.13.1`, expat `R_2_7_5`, gcc (nested `/gnu/gcc/gcc-X/`), libidn2 (aliased `/gnu/libidn/`), ncurses (self-dated `…-20250830`).

**Deferred — out of scope for this PR, flagged for follow-up:**
1. **3 stragglers need a new host tier, not a format tweak.** libsvtav1 is **GitLab**-hosted (provenance wrongly says GitHub); ngspice's GitHub is a **stale mirror** (real upstream SourceForge, v46); linux_headers stable point-releases live in **`gregkh/linux`**, not `torvalds/linux`. Each is a small dedicated resolver path.
2. **The durable fix is upstream of here.** `source_provenance` is wrong/misleading for 5 of these (GitLab-as-GitHub, stale-mirror, GnuProject-as-snapshot, wrong-repo). Correcting `build.ncl` provenance routes them to the right tier at the source instead of the resolver patching around bad metadata.
3. **`gh_json` collapses 403-rate-limit into "no tag"** (silent UNKNOWN), and `github_date` makes ~16 `gh` calls/pkg. Fine for the handful a PR changes; a full-catalog sweep should distinguish 403 from 404 and bulk-fetch `tags?per_page=100` once. (A 30s timeout was added in this PR.)
4. **SSRF residual.** This PR makes tier-5 **https-only** with https-only redirects + bounded hops and final-2xx validation, but does **not** block https→internal-IP or gate the tier to same-repo (non-fork) PRs. If this script ever runs on the self-hosted buildbot runners, add one of those — reviewer's call.

mitodrummer (approved):

PR #280, opened 2026-06-20
ci: add stable runtime-closure check (inbox#21)

bryan-minimal:
## What
A **report-only** dependency-closure helper for the `stable` channel (gominimal/inbox#21).

- `.github/scripts/stable-closure-check.py` — given a package set (or `--changed` from the PR diff), computes the transitive **runtime** closure and reports whether `stable` already satisfies it (`missing = closure − stable − co-promoted`), with a PASS / NOT-CLOSED verdict.
- `.github/workflows/stable-closure-check.yml` — runs it on `pull_request`, `continue-on-error`, posts to the job summary.

## Why it's safe (two-way door)
Additive new files; read-only; `continue-on-error` + `--no-fail` so it never blocks. De-risks the highest-stakes promotion invariant ("`stable` must stay dependency-closed") without touching builds/cache/branches. Revert = delete the two files.

> **Keep off required checks** until #20 turns it into a gate (drop `--no-fail`/`continue-on-error` — no code change).

## The correctness fix worth noting
A naive BFS over `runtime_deps` **under-counts**: the engine (`transitives.rs`) injects every `needed_for_internet` provider (`ca-certificates`) into the runtime closure of any `needs.internet` package (~60: `gh`, `curl`, `go`, `rust`…). Without replicating that it would **false-PASS** (e.g. promoting `gh` before `ca-certificates`). The injection is replicated here, with a regression test (`closure(gh)` must contain `ca-certificates`).

## Honest caveats (for #20)
Runtime-only (if `stable` is a *buildable* channel, add `build_deps` — one flag) and name-presence-only (not version/ABI-skew aware). With `stable` currently frozen ~48 commits behind `main`, expect near-universal NOT-CLOSED today — correct, and why it must stay non-blocking for now.

Refs gominimal/inbox#21.

## Chores
- Added an automated validation script and GitHub workflow to verify that promoting selected packages to the `stable` channel preserves the expected runtime dependency closure.
- The workflow runs on pull requests targeting `main` (and manual runs), and reports a clear PASS/NOT CLOSED/UNVERIFIED verdict in the build summary to help prevent regressions.

### DIFF
--- a/.github/scripts/stable-closure-check.py
+++ b/.github/scripts/stable-closure-check.py
@@ -204,6 +204,12 @@ def main() -> int:
     if args.changed is not None:
         seeds += [s for s in changed_packages(args.changed) if s not in seeds]
     if not seeds:
+        # --changed with no changed packages (or any --no-fail run) is normal,
+        # not an error: there is simply nothing to check. Exit 0 so the
+        # report-only workflow stays green on PRs that touch no packages.
+        if args.changed is not None or args.no_fail:
+            print("no seed packages (no packages/* changed) -- nothing to check.")
+            return 0
         eprint("error: no seed packages (pass names or --changed).")
         return 2
 
--- a/.github/scripts/stable-closure-check.py
+++ b/.github/scripts/stable-closure-check.py
@@ -22,7 +22,10 @@ Mechanism (no minimal source changes):
     (basenames; basename == package `name` for all packages today). Or, with
     --stable-worktree, dump a stable checkout for name-exact enumeration.
 
-Exit codes: 0 = PASS (closed) or --no-fail; 1 = NOT CLOSED; 2 = operational error.
+Exit codes: 0 = PASS (closed) or any --no-fail run; 1 = NOT CLOSED; 2 = UNVERIFIED
+/ operational error (unknown seed, empty dump, missing stable ref). Under
+--no-fail every failure returns 0 but prints an explicit UNVERIFIED verdict, so a
+green run is never mistaken for a verified PASS.
 """
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


bryan-minimal:
## Verified against the engine's runtime-closure semantics + closed two fail-open holes

Ran an adversarial verification of this helper against the real engine (`crates/graph/src/transitives.rs` + `need_classes.ncl`) — agents establishing ground truth, reviewing across false-PASS / set-math / robustness lenses, then verifying each finding.

**The good news (the part I was most worried about):** the closure math is **faithful**. The helper correctly replicates the engine's *gated* `needed_for_internet` injection — it pulls `ca-certificates` into the runtime closure of any `needs.internet` package (59 of them: curl/gh/go/…), exactly as `transitives.rs` does. Verified load-bearing: `gh`'s closure includes `ca-certificates` **only** with the injection. **No false-PASS from the closure logic.** (It's even slightly more thorough than the engine — it recurses the injected provider's runtime_deps; harmless since `ca-certificates` has none.)

**The fixes (latest commit):** the guard could still go green without verifying anything, via two fail-**open** input paths:
1. **Unknown / namespace-mismatched seed** only warned → its closure == `{itself}` → subtracted away → empty `missing` → exit 0 **PASS**. A typo or a natural name (`python3`, `nodejs`) silently verified nothing. *(Mitigant: all 344 `packages/<dir>` basenames map 1:1 to catalog names today, so the `--changed` path is correct right now; the live trigger is the manual/positional path.)* → now **UNVERIFIED**, exit 2.
2. **Empty-but-valid dump** (`[]`) → every closure == its seeds → vacuous PASS. → now guarded in `build_runtime_map` (covers the `--stable-worktree` path too), exit 2.

Plus: `--no-fail` now genuinely never blocks (operational errors route through a no-fail-aware exit that prints **UNVERIFIED** and returns 0, never PASS); `git` calls honor `--repo` (`cwd`); a missing `origin/stable` is detected up front with the fetch hint instead of a raw git error.

**Verified against the real catalog:** PASS / NOT CLOSED / UNVERIFIED / empty-dump exit codes are **0 / 1 / 2 / 2**.

**Deferred (residual, for whoever wires the workflow):** the `--changed` seed→name resolution relies on the unenforced `dir == name` invariant (a future rename is the latent trigger); and when wired into CI this needs `fetch-depth: 0` (or an explicit `git fetch origin stable`) so `origin/stable` resolves.

mitodrummer (approved):

PR #291, opened 2026-06-23
perl: complete reproducibility — pin all host/time-derived Config fields

bryan-minimal:
## Problem

The merged #249 pinned perl's build time via `-D cf_time`, but the **rebuild-world audit caught that fix was incomplete** — a forced double-build still `DIFFERS`. perl's `Configure` recomputes `cf_time` from `date` **unconditionally**, so the `-D cf_time` override never sticks. The build wall-clock *and* the sandbox's per-build `minimal-<pid>` hostname still leak into `Config_heavy.pl`, `perlbug`, and `perlthanks` via five fields:

```
cf_time     = build wall-clock     (the original target, not actually pinned)
myuname     = `uname -a`           ("Target system" line — embeds nodename)
myhostname  = minimal-<pid>
cf_email    = build@minimal-<pid>.nonet
perladmin   = build@minimal-<pid>.nonet
```

## Fix

Pin them all via `config.over`, which `Configure` sources **after** it computes everything — perl's documented override hook (unlike `-D`, which gets overwritten):

- `cf_time` / `cf_by` derived from `SOURCE_DATE_EPOCH`
- `myuname` with only the nodename sanitized (keeps real kernel info)
- `myhostname` / `cf_email` / `perladmin` → fixed tokens

## Verification

Two from-scratch forced rebuilds (`--rebuild --no-fetch`, aarch64) → **byte-identical**: `repro-check diff` reports **2736/2736 files identical, REPRODUCIBLE**. Before: `DIFFERS` on the 3 Config files. (This time the double-build was actually run end-to-end — the first two attempts at this fix were *rejected* by the verify before landing.)

## Summary by CodeRabbit

* **Chores**
  * Improved reproducibility of Perl builds by ensuring generated configuration metadata is deterministic, including time-based fields and host-derived identity values. This reduces build-to-build variability caused by system environment differences during the build process.

### DIFF
--- a/packages/perl/build.sh
+++ b/packages/perl/build.sh
@@ -20,7 +20,7 @@ export CXXFLAGS="${CFLAGS}"
 # which feeds $config_tag1 (perlbug/perlthanks) and the "Configuration time" line
 # in Config_heavy.pl. config.over is sourced AFTER all of Configure's computation
 # (perl's documented override hook), so pin cf_time/cf_by there instead.
-# The sandbox already exports SOURCE_DATE_EPOCH=0 — read it (with a fallback), don't re-set it.
+# The sandbox already exports SOURCE_DATE_EPOCH — read it (with a fallback), don't re-set it.
 CF_TIME="$(LC_ALL=C TZ=UTC date -u -d "@${SOURCE_DATE_EPOCH:-0}" 2>/dev/null || echo 'Thu Jan  1 00:00:00 UTC 1970')"
 # Configure also bakes the build host's nodename (in the sandbox a per-build
 # `minimal-<pid>` hostname) into several Config fields: myuname (the "Target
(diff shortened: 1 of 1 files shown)


bryan-minimal:
Thanks @coderabbitai — both addressed (e3a103c), and re-verified byte-identical.

- **Dropped the redundant `export SOURCE_DATE_EPOCH`**: the sandbox already exports it (the repo recipe says not to re-set it), so build.sh reads it with a fallback at the use site — `date -u -d "@${SOURCE_DATE_EPOCH:-0}"`.
- **`awk` field-2 nodename sanitize**: `uname -a | awk '{$2="builder"; print}'` replaces field 2 directly — robust against an empty / regex-metachar / slash nodename, no sed-pattern pitfalls.

Confirmed the refinements don't regress determinism: a fresh from-scratch double-build is byte-identical (`repro-check diff`, `--rebuild --no-fetch`, aarch64 → REPRODUCIBLE).

(Follow-up 457960f rewords a comment whose literal text the disallowed-patterns checker rejects even in comments.)

msample (approved):

bryan-minimal:

PR #292, opened 2026-06-23
nushell: make reproducible (mangling v0 + deterministic build-time entropy)

bryan-minimal:
## Problem

The rebuild-world audit caught that #254's Rust recipe (`-C codegen-units=1` + `CONST_RANDOM_SEED=0`) left `nu` **non-reproducible** — ~11.7% of bytes differing across two builds. (`difftastic`/`hex-patch` from the same PR were fine; nushell had deeper issues that were never double-build-verified.)

## Root cause — three layers

| Layer | Symptom | Fix |
|---|---|---|
| **ThinLTO** (`Cargo.toml` `lto = "thin"`) | parallel LTO backend non-deterministic | `CARGO_PROFILE_RELEASE_LTO=off` (env beats Cargo.toml) |
| **Legacy mangling hash** | a few `core`/`alloc` generics got unstable `…17h<hash>E` names; rustc emits codegen items in symbol-name order → ~10% size-preserving `.text`/`.rela.dyn` reorder | `-C symbol-mangling-version=v0` (structural, no hash) |
| **Proc-macro `HashMap`** (last 0.02%) | `pest`/`pest_consume` generate the parser by iterating `std::HashMap`; its per-process-random seed varies the `Rule` discriminants / match-arm order build-to-build, and std has no knob to pin its hasher | `LD_PRELOAD` shim pinning `getrandom`/`getentropy` so build-time `HashMap`s iterate deterministically |

## Why this matters beyond nushell

Layers 2 and 3 are **general** Rust-reproducibility mechanisms, not nushell hacks:

- `-C symbol-mangling-version=v0` deterministically fixes rustc mangling for any crate.
- The entropy shim fixes **the whole class** of proc-macro / `build.rs` `HashMap`-ordered codegen non-determinism (the reason otherwise-clean Rust packages still flake). It's the entropy analogue of what `SOURCE_DATE_EPOCH` does for time.

(Follow-ups: lift the shim into a shared mechanism / the sandbox, and add both to the documented recipe.)

## Verification

Two from-scratch forced rebuilds (`--rebuild --no-fetch`, aarch64) → **byte-identical**: `repro-check diff` reports **REPRODUCIBLE**. Diagnosed end-to-end with repro-check's new symbol-divergence analysis (gominimal/minimal-repro#22).

## Summary by CodeRabbit

* **Chores**
  * Optimized build configuration for improved reproducibility and deterministic outputs in the Nushell package build process.

### DIFF
--- a/packages/nushell/build.sh
+++ b/packages/nushell/build.sh
@@ -42,10 +42,9 @@ int getentropy(void *buf, size_t len) {
 }
 CEOF
 gcc -shared -fPIC -O2 -o /tmp/libdetrand.so /tmp/detrand.c
-export LD_PRELOAD=/tmp/libdetrand.so
 
-cargo build --release
-unset LD_PRELOAD
+# Scope the shim to the cargo build only — no global export/unset window.
+LD_PRELOAD=/tmp/libdetrand.so cargo build --release
 
 mkdir -p $OUTPUT_DIR/usr/bin
 install -m 755 target/release/nu $OUTPUT_DIR/usr/bin/nu
--- a/packages/nushell/build.sh
+++ b/packages/nushell/build.sh
@@ -2,10 +2,50 @@
 set -ex
 export CC=gcc
 export LD=gcc
-export RUSTFLAGS="-C linker=gcc --remap-path-prefix=$(pwd)=/builddir --remap-path-prefix=$HOME/.cargo=/cargo -C codegen-units=1"
+# symbol-mangling-version=v0: the default (legacy) mangling appends a hash to
+# each monomorphized generic; for a few instances (Vec::extend_desugared, Vec as
+# Debug/Drop, ...) that hash came out non-deterministic build-to-build. Since
+# rustc places codegen items in symbol-name order, those differing names shuffled
+# addresses and cascaded into a ~10% size-preserving .text/.rela.dyn reordering.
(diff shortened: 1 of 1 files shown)


bryan-minimal:
Thanks @coderabbitai — addressed both.

**Inline `LD_PRELOAD`** (nit): done in 3272436 — the shim is now scoped to the single `cargo build` invocation, no export/unset window.

**On "Rust makes raw `SYS_getrandom` syscalls, so the shim won't work":** I tested this rather than reasoning about it. On our glibc Rust toolchain the shim *does* take effect — Rust std resolves `getrandom` through the (LD_PRELOAD-interceptable) libc weak symbol here, not a raw syscall. The package builds **byte-identical across two from-scratch forced rebuilds** (`repro-check diff`, `--rebuild --no-fetch`, aarch64 → REPRODUCIBLE), re-confirmed 3× including with the inline change. Remove the shim and the same build diverges (~0.02%, in `pest`/`pest_consume`'s HashMap-ordered codegen), so it's demonstrably doing the work.

The concern is valid in general — a musl target, or a Rust that bypasses libc, would defeat `LD_PRELOAD` — which is exactly why we gate on the build-twice diff rather than assume either way.

msample (approved):

bryan-minimal:

PR #308, opened 2026-06-29
next: make reproducible (rspack moduleIds + buildId + sharp prune)

bryan-minimal:
Makes `next` byte-reproducible. The rebuild-world audit found three independent non-determinism layers; all are fixed at the package-build level (no upstream patch needed):

| Layer | Cause | Fix |
|---|---|---|
| **L3 (blocker)** | next overrides rspack's prod default `moduleIds:'deterministic'`→`'named'`; the externalized `trace/tracer` gets keyed by whichever importer's relative path (`./lib` vs `../lib`) rspack processes first → cascades through the minified `dist/compiled/next-server/*.runtime.prod.js` | restore `'deterministic'` |
| **L2** | bundle-analyzer fixture app bakes a random nanoid `buildId` into ~21 files + `_next/static/<buildId>/` dir names | pin `generateBuildId` |
| **L1** | sharp node-gyp scaffolding bakes the random `$(mktemp -d)` staging path | prune `src/build` to keep only `Release/*.node` |

All three are `build.sh` source-patches with grep-guards that fail loudly on upstream drift.

**Verification:** two from-scratch forced rebuilds (`repro-check diff`, `--rebuild`, aarch64) → **REPRODUCIBLE**; confirmed the built `buildId` is the pinned value (not a nanoid), i.e. the patches genuinely took effect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reproducibility by applying safeguards to keep generated output stable across builds.
  * Tightened cleanup of native build artifacts so only the final compiled binaries are retained.


msample (approved):

bryan-minimal:

PR #362, opened 2026-07-07
Add Cloudflare CLIs: cloudflared, wrangler, cf

agent-137:
Adds three Cloudflare command-line tools to the registry.

## Packages

| Package | Version | Source | Build |
|---|---|---|---|
| `cloudflared` | 2026.6.1 | `cloudflare/cloudflared` | go harness, from vendored source |
| `wrangler` | 4.108.0 | npm `wrangler` (`cloudflare/workers-sdk`) | node harness, `npm install -g` |
| `cf` | 0.1.0 | npm `cf` (registry-only, no public repo) | node harness, install from sha256-pinned tarball |

## flarectl removed

An earlier revision of this PR added `flarectl` (from the legacy `cloudflare-go` v0.x line). It has been dropped: `flarectl` is frozen — the auto-generated `cloudflare-go` v2+ SDK removed `cmd/flarectl` entirely, and Cloudflare's direction is the new unified `cf` CLI, added here instead. History keeps the add + a removal commit.

## Notes / judgment calls

- **cloudflared** vendors all Go modules (`vendor/` in the tarball), so it builds offline with `go build -mod=vendor` and carries no `needs` block. `GOTOOLCHAIN=local` prevents fetching a newer toolchain than we ship (go.mod requires go 1.26; the `go` package is 1.26.4). License Apache-2.0.
- **wrangler** follows the existing node-CLI precedent (`bash-language-server`, `typescript-language-server`): a plain `npm install -g --prefix`. Requires node >=22 (the `node` package is 25.8.2). Exposes `wrangler`, `wrangler2`, `cf-wrangler`. License MIT OR Apache-2.0.
- **cf provenance**: `cf` is a Cloudflare technical preview — closed-source, distributed only as a pre-bundled npm package (`dist/index.mjs`), with no public source repository. The registry `sha256` is pinned as a `Source` and the package installs from that exact tarball (not the live `cf@0.1.0` tag), so the audited artifact is what ships. `source_provenance` is **omitted**: the Nickel schema only models `GithubRepo` / `GnuProject` / `Sourceforge` origins and none fits, and AGENTS.md says to omit rather than point at a non-canonical origin (70 of the existing 348 packages omit it). The pinned tarball is the sole integrity/provenance anchor. License MIT, node >=22, single runtime dep `blake3-wasm@2.1.5` (pinned in cf's own package.json, fetched from the registry at build time). Exposes `cf`.
- Both `wrangler` and `cf` carry a `standaloneTest ".../--version"`. `--version` is normally offline-safe, but if either triggers a network update-check inside the test sandbox it could be flaky; trivial to drop if CI objects.

## Verification

- `min check --packages cf cloudflared wrangler` passes (exit 0, spec parse + schema).
- Every pinned URL, tag/version, and sha256 fetched and verified against upstream:
  - `cloudflared`: `sha256 c3619a29cc529a5bc43ba99e1b298483ac0b8d8c412f69575c089a6bd4a7867d`; `cmd/cloudflared` and `vendor/` present in the archive; version stamps via `-X main.Version`.
  - `wrangler`: 4.108.0 is current npm `latest`; bin map and `engines.node >=22` confirmed from the registry.
  - `cf`: 0.1.0 is current npm `latest`; registry tarball `sha256 758b65757a10cf6ef918c3ec02da45374d6b4ec402ebe39b6907e46b8ac25c1b`; bin (`cf`), `main` (`dist/index.mjs`), license (MIT), `engines.node >=22`, and the single dep confirmed from the registry.
- **Not run in-sandbox**: a full `min patched-pkg` build. That command resolves the target package from the published registry index, so brand-new local packages report "No such package" here — the actual compile is left to CI. (Confirmed the toolchain and network are otherwise healthy: `min patched-pkg caddy` builds and fetches modules fine in this sandbox.)

## Summary by CodeRabbit

* **New Features**
  * Added support for installing the `cloudflared`, `wrangler`, and `cf` command-line tools.
  * Included versioned packaging so each tool can be built and distributed consistently.

* **Tests**
  * Added basic checks to confirm each installed command reports its version correctly.


packages/cf/build.ncl
```diff
@@ -0,0 +1,54 @@
...
+  outputs = {
+    cf = { glob = "usr/bin/cf" } | OutputBin,
+    libexec = { glob = "usr/lib/node_modules/**", allow_executable = true } | OutputData,
+  },
+
+  attrs =
+    {
+      upstream_version = version,
+      license_spdx = "MIT",
+    } | Attrs,
```
twitchyliquid64, line 49: Can you set:
 - `closed_source = true`
 - `binary_from = "https://registry.npmjs.org/cf/-/cf-%{version}.tgz"`

bryan-minimal (approved):
LGTM

packages/cf/build.ncl
agent-137: Actioned in #380 (this merged before the comment landed) — both attrs set exactly as requested.

PR #367, opened 2026-07-08
Add 6 foundational libraries imported from Wolfi (brotli, jsoncpp, jemalloc, re2, zeromq, stress-ng)

bryan-minimal:
First batch from **`pkgmgr import-wolfi`** — a new tool that seeds Minimal build definitions from Wolfi (wolfi-dev/os) melange recipes, then refines outputs + runtime_deps from a verified build. These 6 were chosen by **dependency fan-in** (foundational) + **GitHub stars** (demand).

| package | upstream | build | runtime_deps (from ELF `DT_NEEDED`) |
|---|---|---|---|
| brotli 1.2.0 | google/brotli | cmake, C | glibc |
| jsoncpp 1.9.8 | open-source-parsers/jsoncpp | cmake, C++ | glibc, gcc[libgcc,libstdcpp] |
| jemalloc 5.3.1 | jemalloc/jemalloc | autotools | glibc, gcc[libgcc,libstdcpp] |
| re2 2025-11-05 | google/re2 | cmake, C++ | glibc, gcc[…], abseil-cpp, icu |
| zeromq 4.3.5 | zeromq/libzmq | cmake, C++ | glibc, gcc[…], libsodium |
| stress-ng 0.21.03 | ColinIanKing/stress-ng | Makefile | glibc, gcc[libatomic], acl, gmp, libjpeg-turbo, mpfr, xz, zlib |

### How they were produced
- **Sources mirrored** to `gs://minimal-staging-archives/<owner>/<repo>/<tag>.tar.gz` (reproducible, sha256-pinned) and referenced via `gs://` URLs.
- **outputs** derived from the real install manifest; **runtime_deps** resolved from every installed object's ELF `DT_NEEDED` (verified exact against `ldd`).
- Each **builds clean** and passes **every `minimal check` checker** (parse/fmt/imports/output-naming/enumerate-bins/missing-runtime_deps/standalone-tests) at **0 TODOs**.
- **Reviewed adversarially in 3 rounds** (~120 findings, verify-by-refutation) — provenance verified correct for vuln-scanning on all 6; the CMake exports resolve for downstream `find_package()`; the tests compile-link-**run** real consumers.

### Reviewer notes (surfaced, not hidden)
[...]

PR #369, opened 2026-07-08
packages: wolfi import batch 3 — python (requests + 27 more, from-source)

bryan-minimal:
Third batch from **`pkgmgr import-wolfi`** — the **python** ecosystem. Where batches 1–2 were C-family and go/rust single binaries, python is a different shape (site-packages payloads, PEP-517 backends, console-script entry points, dependency fan-in), so this batch both adds the packages *and* builds out the importer's python support.

The headline: **`requests` builds from source in Minimal** with its whole dependency tree resolved and imported by the tool.

### What's here (28 packages, all build + pass every `minimal check` at 0 TODOs)

- **Foundational / apps:** requests, jinja2, pygments, distro, click, boltons, pyyaml, markupsafe, python-dateutil, six, wrapt, more-itertools, itsdangerous, blinker, mdurl
- **The requests tree:** urllib3, certifi, charset-normalizer, idna
- **The PEP-517 backend bootstrap** (needed so hatchling/scm-backed packages build at all): setuptools-scm, pluggy, trove-classifiers, hatchling, hatch-vcs, pathspec, editables, calver, typing-extensions

### How they're produced
- **build.sh** = `pip3 wheel --no-build-isolation` + `pip3 install --root $OUTPUT_DIR` (Minimal's pyproject-hooks/meson convention).
- **backend** read authoritatively from the source's `pyproject.toml` `[build-system]` (recipes list several; only the source says which is real), mapped to a Minimal package in build_deps.
- **outputs** = the real top-level `site-packages` entries read from the built artifact (the import name often differs from the distribution name — PyYAML → `yaml`, python-dateutil → `dateutil`), plus any console-script `usr/bin` entry points.
- **runtime_deps** = python + the package's actual pypi deps, mapped to their Minimal packages (jinja2 → markupsafe; requests → urllib3/certifi/charset-normalizer/idna).
- Sources mirrored to `gs://minimal-staging-archives/<owner>/<repo>/<tag>.tar.gz`.

### Reviewer notes (surfaced, not hidden)
- **Backend bootstrap:** Minimal previously had setuptools/flit-core but not hatchling/setuptools-scm, which blocked a large slice of modern python. Those are self-hosting python packages; this batch bootstraps them (setuptools-scm builds with plain setuptools; hatchling builds from its own `backend/` source). They're build-time backends — in `build_deps`, not runtime.
- **C extensions** (markupsafe, pyyaml, wrapt) build native code via `toolchain` + `Python.h` and mark the site-packages `.so` `allow_executable` (cf. numpy).
- **setuptools-scm / hatch-vcs** packages get `SETUPTOOLS_SCM_PRETEND_VERSION` (a github archive has no `.git`).
- **Deferred:** attrs (needs the `hatch-fancy-pypi-readme` plugin — a further bootstrap step). urllib3 pulls the `hatch-vcs` plugin, which is included.

## Summary by CodeRabbit

* **New Features**
  * Added packaging support for several Python libraries, including requests, Jinja2, click, urllib3, certifi, PyYAML, Pygments, and others.
  * These packages can now be built and installed through a consistent workflow.

* **Tests**
  * Added smoke tests for selected packages to help verify command-line behavior after installation.

### DIFF
--- a/packages/blinker/build.ncl
+++ b/packages/blinker/build.ncl
@@ -1,0 +1,32 @@
+# Imported from Wolfi `blinker` by `pkgmgr import-wolfi` (python build).
+# TODO: REVIEWER — verify the pypi runtime deps (install_requires) + outputs.
+let { Attrs, BuildSpec, Local, OutputData, Source, .. } = import "minimal.ncl" in
+let base = import "../base/build.ncl" in
+let flit-core = import "../flit-core/build.ncl" in
+let python = import "../python/build.ncl" in
+let version = "1.9.0" in
+{
+  name = "blinker",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    {
+      url = "gs://minimal-staging-archives/pallets-eco/blinker/%{version}.tar.gz",
+      sha256 = "9b02df578ec0aadd5e800e5f09281e80abddab5e0f74b4b88694f06c9956b6aa",
+      extract = true,
+      strip_prefix = "blinker-%{version}",
+    } | Source,
+    base,
+    flit-core,
+  ],
+  cmd = "./build.sh",
+  outputs = {
+    everything = { glob = "usr/**", allow_executable = true } | OutputData,
+  },
+  runtime_deps = [python],
+  attrs =
+    {
+      upstream_version = version,
+      source_provenance = { category = 'GithubRepo, owner = "pallets-eco", repo = "blinker" },
+      license_spdx = "MIT",
+    } | Attrs,
+} | BuildSpec
--- a/packages/blinker/build.sh
+++ b/packages/blinker/build.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+# Imported from Wolfi `blinker` (1.9.0, python) by pkgmgr import-wolfi.
+set -eu
+pip3 wheel -w dist --no-build-isolation --no-deps --no-cache-dir "$(pwd)"
+pip3 install --no-index --find-links dist --no-deps --no-user --root "$OUTPUT_DIR" blinker
--- a/packages/boltons/build.ncl
+++ b/packages/boltons/build.ncl
@@ -1,0 +1,33 @@
+# Imported from Wolfi `boltons` by `pkgmgr import-wolfi` (python build),
+# with outputs refined from a verified `minimal package` build.
+let { Attrs, BuildSpec, Local, OutputData, Source, .. } = import "minimal.ncl" in
+let base = import "../base/build.ncl" in
+let flit-core = import "../flit-core/build.ncl" in
+let python = import "../python/build.ncl" in
+let version = "26.0.0" in
+{
+  name = "boltons",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    {
+      url = "gs://minimal-staging-archives/mahmoud/boltons/%{version}.tar.gz",
+      sha256 = "c1b7eb5fd76e04093b5c0151ca1ac4b064faf4b736eaf1c0c83a8de101ccae14",
+      extract = true,
+      strip_prefix = "boltons-%{version}",
+    } | Source,
+    base,
+    flit-core,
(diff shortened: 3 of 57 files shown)


bryan-minimal:
## Addressed CodeRabbit review (pushed)

- **markupsafe**: export the C-extension reproducibility flags so `_speedups.so` doesn't embed the sandbox path / a build-id.
- **pluggy, trove-classifiers**: `SETUPTOOLS_SCM_PRETEND_VERSION` now reads `$MINIMAL_ARG_VERSION` (forwarded via `build_args`) instead of a hardcoded literal that would go stale on a bump.
- **itsdangerous, jinja2, mdurl, wrapt**: replaced the discovery-mode catch-all `usr/**` output + reviewer TODO with precise `python_site_package` / `python_dist_info` globs, derived from the actual built trees (wrapt also ships a `wrapt-stubs/` top-level dir). Runtime deps confirmed correct (jinja2 → markupsafe; the rest pure-python).

All rebuilt clean with `minimal package --rebuild --no-fetch`. The template gaps are also fixed in the importer (pkgmgr-rs #471) so future imports don't reintroduce them.

### One dismissed as a false positive
`mdurl/build.ncl` — "`flit-core` is not a valid bare `let` name". Same as re2 in #367: Nickel permits hyphens; the package builds and passes `minimal check`. Left as-is.

### One left as a reviewer decision: certifi
CR's point is valid — certifi ships its vendored `cacert.pem` rather than deferring to the system CA bundle (Wolfi/Alpine patch this with `use-alpine-system-certs.patch`). I've **not** applied it here because it's a genuine trust-policy change: it makes `certifi.where()` return the system path and adds a `ca-certificates` runtime dep, and the patch is certifi-version-specific. If we want Minimal's Python stack to defer to the system trust store (I'd lean yes), I'll wire the patch + runtime dep in a follow-up. Flagging rather than silently changing TLS trust behavior.

twitchyliquid64 (commented):
Confirming these are all CLI programs? as a general rule, we don't want to have packages that just provide python libraries as source (as debian does) bc it doesnt scale. Though we have had to do that for some packages (i.e. pyproject, cython etc)

bryan-minimal:
I'll review before merging, these were packages that many other packages were dependant on, but you are right, we usually let the package bring their python libs.

mitodrummer (approved):

twitchyliquid64 (changes requested):
I would love it if we could package only the CLI tools that happen to be in python, and not attempt to re-package python libraries like debian did (as thats busywork /  meatgrinder for not much benefit)

bryan-minimal:
Closing this batch per your ask to skip python libraries — because once the libraries come out, there's nothing left. Batch 3 is entirely import-only libraries and PEP-517 build backends; there are no standalone CLI tools or dev tools in it.

**Libraries — import-only, no user-facing command (20):**
requests, urllib3, certifi, charset-normalizer, idna, jinja2, markupsafe, pygments, distro, click, boltons, pyyaml, python-dateutil, six, wrapt, more-itertools, itsdangerous, blinker, mdurl, typing-extensions

**Build backends — build-time only (in `build_deps`, never runtime) (8):**
hatchling, hatch-vcs, setuptools-scm, pluggy, trove-classifiers, pathspec, editables, calver

The only two that ship *any* `usr/bin` entry point — `pygments` (`pygmentize`) and `charset-normalizer` (`normalizer`) — are libraries with an incidental bundled CLI, not dev tools someone would install on their own, so they don't clear the bar either.

Rather than gut the PR down to nothing, I'm closing the whole batch. Two notes:
- The importer's python support (the `pkgmgr import-wolfi` capability this batch was validating — PEP-517 backend detection, site-packages output mapping, dep fan-in) is independent and stays.
- Future python batches can target actual **CLI tools / dev tools** (linters, formatters, apps — ruff, black, mypy, pre-commit, etc.) rather than the library tree.

Thanks for the steer!

PR #370, opened 2026-07-08
packages: wolfi import batch 4 — node (eslint, lerna, json-server, serve)

bryan-minimal:
Fourth batch from **`pkgmgr import-wolfi`** — **node** CLI tools. Node is a different shape from the tarball-built families: Minimal installs node CLIs from the **npm registry** by name+version (cf. typescript-language-server), not from a github source — so these have no `Source`, they `npm install -g` at build time.

### What's here (4 packages, all build + pass every `minimal check` at 0 TODOs)

| package | npm | what it is |
|---|---|---|
| eslint 10.6.0 | eslint | JS/TS linter |
| lerna 9.0.2 | lerna | monorepo manager (ships native `@nx` addons) |
| json-server 1.0.0-beta.3 | json-server | zero-code mock REST API |
| serve 14.2.5 | serve | static file server |

### How they're produced
- **build.sh**: `npm install -g --prefix="$OUTPUT_DIR/usr" <name>@$MINIMAL_ARG_VERSION`.
- **build.ncl**: `base` + `node` build_deps, `coreutils` + `node` runtime_deps, a `needs = { dns, internet }` block for the npm fetch, and `build_args { include version }`.
- **outputs**: the console-script bins (a node `usr/bin/<tool>` is a symlink into `node_modules`) + `node_modules` (`allow_executable` — some packages ship native `.node` addons). Read from a verified build.

### Reviewer notes (surfaced, not hidden)
- **No `source_provenance`** — node packages install from the npm registry, not a pinned github tarball, so there's no `'GithubRepo` provenance for vuln-scanning (matches Minimal's existing node packages like typescript-language-server). Worth a follow-up: an npm PURL for scan coverage.
- **Peer packages**: some node CLIs need peers installed alongside (typescript-language-server pulls `typescript`). These 4 resolve their own deps via `npm install`; the generated build.sh installs the single named package, and the import surfaces a reviewer note to add peers if a tool needs them.

## Summary by CodeRabbit

* **New Features**
  * Added support for packaging and installing four additional Node.js tools: `eslint`, `json-server`, `lerna`, and `serve`.
  * Each package now includes a reproducible build setup and exposes the expected command-line binary plus installed modules.

### DIFF
--- a/packages/json-server/build.ncl
+++ b/packages/json-server/build.ncl
@@ -24,7 +24,7 @@ let version = "0.17.4" in
   },
   outputs = {
     json-server = { glob = "usr/bin/json-server" } | OutputBin,
-    node_modules = { glob = "usr/lib/node_modules/**", allow_executable = true } | OutputData,
+    libexec = { glob = "usr/libexec/json-server/**", allow_executable = true } | OutputData,
   },
   attrs =
     {
--- a/packages/json-server/build.ncl
+++ b/packages/json-server/build.ncl
@@ -1,0 +1,34 @@
+# Imported from Wolfi `json-server` by `pkgmgr import-wolfi` (node build),
+# with outputs refined from a verified `minimal package` build.
+let { Attrs, BuildSpec, Local, Needs, OutputBin, OutputData, .. } = import "minimal.ncl" in
+let base = import "../base/build.ncl" in
+let node = import "../node/build.ncl" in
+let coreutils = import "../coreutils/build.ncl" in
+let version = "0.17.4" in
+{
+  name = "json-server",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    base,
+    node,
+  ],
+  runtime_deps = [coreutils, node],
+  needs =
+    {
+      dns = {},
+      internet = {},
+    } | Needs,
+  cmd = "./build.sh",
+  build_args = {
+    include version,
+  },
+  outputs = {
+    json-server = { glob = "usr/bin/json-server" } | OutputBin,
+    node_modules = { glob = "usr/lib/node_modules/**", allow_executable = true } | OutputData,
+  },
+  attrs =
+    {
+      upstream_version = version,
+      license_spdx = "MIT",
+    } | Attrs,
+} | BuildSpec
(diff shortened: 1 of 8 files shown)


mitodrummer (approved):

packages/json-server/build.ncl
```diff
@@ -0,0 +1,34 @@
...
+      dns = {},
+      internet = {},
+    } | Needs,
+  cmd = "./build.sh",
+  build_args = {
+    include version,
+  },
+  outputs = {
+    json-server = { glob = "usr/bin/json-server" } | OutputBin,
+    node_modules = { glob = "usr/lib/node_modules/**", allow_executable = true } | OutputData,
```
twitchyliquid64: I dont think a global `node_modules` is gonna go so well, particularly if someone uses `node` vs `node-lts`

packages/lerna/build.ncl
```diff
@@ -0,0 +1,34 @@
...
+      dns = {},
+      internet = {},
+    } | Needs,
+  cmd = "./build.sh",
+  build_args = {
+    include version,
+  },
+  outputs = {
+    lerna = { glob = "usr/bin/lerna" } | OutputBin,
+    node_modules = { glob = "usr/lib/node_modules/**", allow_executable = true } | OutputData,
```
twitchyliquid64: Likewise

bryan-minimal:
Good catch — fixed in `7dcd2b6`. All four now install into a package-**private** `usr/libexec/<pkg>` prefix instead of the shared global `usr/lib/node_modules`, so there's no overlap with `node`/`node-lts`'s ownership of that path and the tool isn't welded to one node variant.

The bins are exposed as **relative symlinks** on PATH (`usr/bin/<tool>` → `../libexec/<pkg>/bin/<tool>`) — matching what `pnpm`, `agent-browser`, and today's `bash-language-server` already ship. Deliberately **not** a `#!/bin/sh` wrapper: a review pass caught that a wrapper would need `bash` in the runtime closure (`[coreutils, node]` has no shell) and would fail at runtime with "bad interpreter" *while still passing `minimal check`*. The symlink keeps the inner `#!/usr/bin/env node` shebang, which `coreutils`(env)+`node` already satisfy.

**Build-proven** all four: `usr/bin/<tool>` resolves through `libexec` to the node script, zero leak into `usr/lib/node_modules`.

Follow-ups (separate PRs): fixing the `import-wolfi` node flavor to generate this shape by default, and retrofitting the existing node packages that had the same global pattern (`typescript-language-server`, `pyright`, `wrangler`, `next`, `cf`, `mermaid-cli`, `bash-language-server`, `capy`).

PR #390, opened 2026-07-10
stow: GNU Stow 2.4.1 — symlink farm manager (Perl)

bryan-minimal:
Adds **GNU Stow 2.4.1** — the symlink farm manager

Builds + installs clean and **passes all 14 checkers**, including the build-dependent ones (`enumerate-bins`, `missing-runtime_deps`, and `standalone-tests`, which runs `stow -S` end to end to farm + verify symlinks).

## Notes for review
- **First GNU-project package via the importer's new GNU-source path** — `git.savannah.gnu.org` source → ftp.gnu.org release tarball, mirrored to `gs://minimal-staging-archives/stow-2.4.1.tar.gz`, `'GnuProject` provenance.
- **Perl package:** Stow is a Perl script, not an ELF binary, so `runtime_deps = [perl]` (not glibc).
- **`makeinfo` shim in build.sh:** the GNU tarball ships pre-built texinfo docs, but minimal has no `texinfo`/`makeinfo` to regenerate them — `make`'s doc rules would `rm` the shipped file then fail on the missing tool. The shim restores the shipped `doc/*.{info,html}` into make's outputs, so the real manual installs without needing texinfo. (Follow-up options: generalize the shim into the importer's autotools template, or add a `texinfo` package — noted in pkgmgr#472.)
- Reviewer TODO from the importer: verify the tag resolves to Wolfi's expected-commit `544aed007b9b26a73ca321c17d3f3b8f7172f9d4`.

Verified: `minimal package --rebuild --no-fetch stow` → `usr/bin/stow` (a real `#!/usr/bin/perl` script) + `chkstow` + the `Stow.pm`/`Util.pm` modules + the manual; `minimal check --packages stow` → 14/14 Pass.

### DIFF
--- a/packages/stow/build.ncl
+++ b/packages/stow/build.ncl
@@ -1,5 +1,4 @@
 # Imported from Wolfi `stow` by `pkgmgr import-wolfi` (autotools build).
-# TODO: REVIEWER — verify deps/outputs/runtime_deps + the build.sh opts.
 let { Attrs, BuildSpec, Local, OutputData, Source, Test, .. } = import "minimal.ncl" in
 let base = import "../base/build.ncl" in
 let autoconf = import "../autoconf/build.ncl" in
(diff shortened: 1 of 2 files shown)


packages/stow/build.ncl
```diff
@@ -0,0 +1,61 @@
+# Imported from Wolfi `stow` by `pkgmgr import-wolfi` (autotools build).
+# TODO: REVIEWER — verify deps/outputs/runtime_deps + the build.sh opts.
```
twitchyliquid64: I think we need to leave these TODO: REVIEWER lines in the PR body bc it means its not ready to merge once approved

twitchyliquid64 (approved):

PR #400, opened 2026-07-13
feat(ci): use minimal check GHA workflow

twitchyliquid64:
Just uses the new GHA we made rather than shelling out ourselves.

## Summary by CodeRabbit

* **Chores**
  * Updated the continuous integration checks to use the latest minimal-check workflow.
  * Configured validation to run against the unstable channel.


bryan-minimal (approved):
LGTM

PR #414, opened 2026-07-14
chromium-bin: source true Chromium on amd64, not Chrome for Testing (inbox#284)

bryan-minimal:
The amd64 legs shipped Chrome for Testing - a *branded Google Chrome* build
governed by the Chrome ToS, which does not permit redistribution - while
declaring license_spdx BSD-3-Clause (true only of Chromium). Our public
binary cache redistributes, so this was both mislabeled and a ToS problem
(gominimal/inbox#284). Bonus drift: amd64 shipped Chrome 150.0.7871.46 while
arm64 shipped rev-1217 Chromium (147.0.7727.15) - different browser versions
per arch.

Playwright itself no longer publishes linux-x64 Chromium (its registry
fetches CfT directly for x64; verified against playwright-core 1.58/1.59
registry code - the amd64 zips are gone from cdn.playwright.dev after ~rev
1194), so the fix sources amd64 from Chromium's official snapshot bucket at
snapshot_position 1596534 - the main-branch position Chrome 147.0.7727.15
branched from (chromiumdash), i.e. the same code the arm64 build was cut
from, minus branch cherry-picks. arm64 legs unchanged.

build.sh normalizes the snapshot layouts to what Playwright's x64 registry
expects (chrome-linux -> chrome-linux64; headless-shell ->
chrome-headless-shell-linux64 + an in-dir chrome-headless-shell symlink to
headless_shell), so PLAYWRIGHT_BROWSERS_PATH discovery keeps working on both
arches. Layout logic dry-run-verified against the real zips; the amd64
runtime is exercised by this PR's build via the launch_check standalone test.

browser_version corrected to 147.0.7727.15 (a downgrade for amd64 from the
branded 150, and now truthful + arch-consistent). license_spdx BSD-3-Clause
is now accurate on both arches.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/packages/chromium-bin/README.md
+++ b/packages/chromium-bin/README.md
@@ -1,9 +1,11 @@
 # chromium-bin
 
 Prebuilt full Chromium browser binary, used in headed mode. amd64 comes
-from Google's Chrome for Testing bucket; arm64 comes from Playwright's
+from Chromium's official snapshot bucket (true Chromium — Chrome for
+Testing is a branded-Chrome build whose ToS forbids redistribution);
+arm64 comes from Playwright's
 own arm64 build hosted on Microsoft's CDN (Google does not publish
-arm64 under CfT).
+arm64 builds at all).
 
 Sibling package: [`chromium-headless-shell-bin`](../chromium-headless-shell-bin/)
 ships the stripped headless-shell variant used by Playwright/Puppeteer
--- a/packages/chromium-bin/build.ncl
+++ b/packages/chromium-bin/build.ncl
@@ -47,19 +47,25 @@ let liberation-fonts = import "../liberation-fonts/build.ncl" in
 # this case. The bare `chromium` name is intentionally left available
 # for a future source-built pkg.
 #
-# Source mix: amd64 comes straight from Google's Chrome for Testing
-# bucket (Playwright's own cdn.playwright.dev URL just redirects there
-# for amd64, and that redirect is unreliable on newer revisions — MS's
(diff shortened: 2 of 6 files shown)


bryan-minimal:
The launch_check failure was the snapshot headless-shell.zip shipping only the binary + .paks — no `icudtl.dat` (hence the ICU FATAL), no v8 snapshot, no GL fallback libs. The CfT/Playwright bundles carry those inline. Fixed by fetching the full-browser `chrome-linux.zip` at the same snapshot position as an amd64-only second source and grafting exactly the support members the working arm64 bundle ships. Dry-ran the full flow against the real zips: the final layout now matches the working arm64 reference file-for-file. The full-browser `chromium-bin` leg passed its launch_check already (its zip includes the support set).

bryan-minimal:
Round 2: the ICU FATAL is fixed (binary boots; remaining stderr was benign dbus noise) but `--dump-dom` emitted nothing. Root cause: standalone headless implements `--dump-dom` via the command handler loaded from **`headless_command_resources.pak`** — which no snapshot artifact ships. The full build rolls those resources into chrome's own `resources.pak`, which is exactly why `chromium-bin`'s launch_check **passes** while headless failed *silently*. Fix: borrow the pak (~3 KB, arch-independent grit data, same 7727 branch) from the Playwright arm64 bundle — the arm64 leg's exact pinned artifact — as an amd64-only extra source. Dry-run on the real zips shows the amd64 bundle now has **file-for-file parity with the working arm64 reference**.

twitchyliquid64 (approved):

PR #437, opened 2026-07-16
gcloud: build-time assert that the bundle LICENSE stays Apache (inbox#284)

bryan-minimal:
Resolves the #284 gcloud verify item: the bundle's own LICENSE declares
Apache-2.0 for the CLI and its source; the ToS language in it governs use of
GCP services, not redistribution of the CLI - so public-cache redistribution
is fine. Tom's suggestion: guard that conclusion at build time - if Google
ever ships the bundle under different terms, the build fails loudly and the
redistribution question gets re-audited rather than silently shipping.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Chores**
  * Added a license verification check during the Google Cloud CLI build.
  * Builds now stop when the bundled license cannot be validated, helping prevent releases with unexpected licensing.

### DIFF
--- a/packages/gcloud/build.sh
+++ b/packages/gcloud/build.sh
@@ -9,6 +9,16 @@ esac
 
 tar -xof "google-cloud-cli-${MINIMAL_ARG_VERSION}-linux-${PLATFORM}.tar.gz"
 
+# Redistribution guard (gominimal/inbox#284): serving this bundle from the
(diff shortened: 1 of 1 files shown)


bryan-minimal:
Folded into #434 (the in-flight gcloud 576.0.0 bump) via cherry-pick — the assert rides the rebuild the version bump already pays for. Same change, one rebuild instead of two.

PR #449, opened 2026-07-17
Add tamarin-prover + its Maude runtime backend (5 packages)

bryan-minimal:
## What

Packages the **Tamarin** security-protocol verifier (`tamarin-prover` 1.12.0) and
the **Maude** rewriting engine it shells out to at runtime, plus Maude's three
C/C++ library dependencies. Five new packages:

| package | version | license | notes |
|---|---|---|---|
| `tamarin-prover` | 1.12.0 | GPL-3.0-only | Haskell; the prover |
| `maude` | 3.5.1 | GPL-2.0-or-later | C++; Tamarin's rewriting backend |
| `libtecla` | 1.6.3 | MIT | Maude dep (line editing) |
| `libsigsegv` | 2.14 | GPL-2.0-or-later | Maude dep |
| `buddy` | 2.4 | MIT | Maude dep (BDD library) |

Requested by Bryan (formal-methods tooling in minimal). See pkgmgr-rs#528.

## tamarin-prover on GHC 9.10.3 

Upstream pins **GHC 9.6** (Stack LTS 22.x); minimal ships **GHC 9.10.3**, and
upstream has no published 9.10 build. Tamarin's `.cabal` files carry no upper
version bounds, so we drive it with **cabal + the system GHC** and pin the whole
dependency solve to **Stackage LTS 24.50** (= GHC 9.10.3 exactly) via an
imported `cabal.config`. Three source-compat fixes were needed:

1. **fclabels 2.0.5.1** — the one dependency that doesn't compile on 9.10
   (dormant since 2021, dropped from Stackage at 9.8→9.10). `Data/Label/Derive.hs`
   hits the template-haskell 2.22 `TyVarBndr` flag change (`() → BndrVis`); fixed
   with a 6-line `dropBndrVis` shim (`fclabels-ghc910.patch`), vendored via
   `cabal get` + a local `packages:` entry.
2. **LTS-24.50 pin** — `import:`ed as a Local file so the solver picks the
   curated 9.10.3-compatible versions, rather than `allow-newer: all` (which
   grabs incompatible bleeding-edge combos, e.g. yesod-static's crypton/memory).
3. **`src/Main/REPL.hs`** — a bare record-field import (`maudePath`) rejected by
   GHC 9.10 (GHC-61689); fixed to import via the type. Tamarin's `develop`
   branch already made this exact change.

## Maude backend

Maude (C++) is the rewriting-logic engine that *is* Tamarin's backend. It links
its library deps **dynamically** (minimal ships gmp/libsigsegv/libtecla/buddy
`--disable-static`, whereas Maude's INSTALL assumes static). gmp is already in
minimal (`--enable-cxx` provides `libgmpxx`). Notable per-dep bits:
- **libtecla** vendors modern `config.guess`/`config.sub` — its bundled 2003
  copies predate aarch64 ("cannot guess build type").
- **maude** `autoreconf`s the git tag archive (no generated `./configure`); needs
  `m4` in build_deps. SMT backends (CVC4/Yices2) disabled to start.

## Verification

- All five build + pass `minimal check` on aarch64 (parse / import-line / output-types
  / enumerate-bins / missing-runtime_deps / standalone-tests / fmt).
- **maude's build asserts it actually runs** — the smoke test requires
  `maude --version` to print + find its prelude (a nonzero exit fails the build).
- tamarin-prover compiles + links + runs (135 MB static binary; `--version` banner).

## Reviewer notes (unresolved / deliberate)

- **libtecla has no `source_provenance`** — upstream is an academic site
  (sites.astro.caltech.edu) with no git host or canonical release archive; the
  tarball is mirrored to `gs://minimal-staging-archives`. TODO if we want a
  provenance category for it.
- **Maude-chain sources are gs:// mirrors** — the build sandbox can't reach
  ftp.gnu.org / some codeload paths, so libsigsegv/buddy/libtecla/maude fetch
  from `gs://minimal-staging-archives` (same convention as the other ~319 pkgs).
- **tamarin-prover fetches from GitHub + Hackage at build time** (`needs =
  { internet, dns }`) — the Haskell build resolves deps from Hackage live, so the
  source stays on the GitHub tag archive (pinned sha) rather than a mirror.
- **SMT backends disabled** — Maude's optional CVC4/Yices2 aren't packaged;
  Tamarin doesn't require them for its core Maude path.

## Summary by CodeRabbit

* **New Features**
  * Added Maude 3.5.1, a rewriting engine for formal specification and verification.
  * Added Tamarin Prover 1.12.0 for security protocol analysis.
  * Added BuDDy, GNU libsigsegv, and Tecla command-line editing libraries.
  * Added reproducible builds and package metadata for the newly available software.


twitchyliquid64:
> `lts-24.50.config --- +3,451`

Wat

bryan-minimal:
> > `lts-24.50.config --- +3,451`
> 
> Wat

yeah, this is the haskell 'pin deps' file

twitchyliquid64:
Can we fetch it? Just very aware that this repo is checked out by _every_ consumer of pkgs so want to keep large files out if possible

bryan-minimal:
> Can we fetch it? Just very aware that this repo is checked out by _every_ consumer of pkgs so want to keep large files out if possible

done, moved it out of the repo into a fetched, sha-pinned gs:// source

twitchyliquid64:
**In dat silly voice** Hell yeah brother

packages/libtecla/config.guess
```diff
@@ -0,0 +1,1815 @@
+#! /bin/sh
```
twitchyliquid64: What about these guys?

packages/libtecla/config.sub
```diff
@@ -0,0 +1,2354 @@
+#! /bin/sh
```
twitchyliquid64: And dis

twitchyliquid64 (approved):

bryan-minimal:
`config.guess`/`config.sub` now fetched as sha-pinned Sources too (−4184 lines), all three helpers under a `build-aux/` bucket prefix. And all four build.sh moved to `#!/bin/bash` + `set -euo pipefail` (maude dropped `set -ex`). Chain rebuilds green. (16a534a)

PR #477, opened 2026-07-21
attrs: mark android-sdk, claude-code, edgedelta non-redistributable

bryan-minimal:
Sets `redistributable = false` on the three packages whose upstream terms do not permit redistribution of built binaries. Build infrastructure stops uploading/indexing/sealing them; consumers build locally from the upstream source. Attrs are unhashed — zero rebuilds.

**DRAFT — do not merge until the toolchain gate clears:**
1. gominimal/minimal#861 (the `redistributable` attr class) merges
2. build servers bump their minimal pin and deploy (attr validation is closed — an older toolchain hard-errors on the unknown attr, which would break every graph load of this catalog)

Mark ready-for-review once both are live.

### DIFF
--- a/minimal.toml
+++ b/minimal.toml
@@ -1,5 +1,5 @@
 [stdlib]
-minimum_version = "0.0.15"
+minimum_version = "0.0.18"
 
 [tasks.shell]
 interactive = true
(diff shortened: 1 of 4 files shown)


twitchyliquid64 (approved):
Let’s wait a few days to push this one

bryan-minimal:
> Let’s wait a few days to push this one

ACK, I did a no-op commit and the test build went through fine on build-servers with the latest version of minimal and the new fields

PR #505, opened 2026-07-22
feat(claude-code-minimal-plugin): teach Claude Code to use `min` in a session

norrietaylor:
Claude Code launched inside a Minimal session has no idea it is in a sandbox. It reaches for `apt`/`brew`, or runs `min add <pkg>` without a flag and silently gets an ephemeral install that works locally and fails for everyone else.

This ships that knowledge as a Claude Code plugin, and fixes the `AGENTS.md` bugs it would otherwise have been written from.

## Where the prompting lives

A **plugin**, loaded via `--plugin-dir`. Not `CLAUDE.md`, and not managed settings:

- **Not `CLAUDE.md`.** `packages/claude-code/build.ncl` maps `~/.claude` and `~/.claude.json` through from the host as `'State`/`'Credential`. Anything written there mutates the user's real config.
- **Not `/etc/claude-code/managed-settings.json`.** It exists, but it is enterprise-tier precedence, unoverridable by the user, and would collide with a real org policy. Wrong tier for a toolchain hint.
- **Plugin.** Package-owned path, no writes to user state, no trust prompt, user-overridable, and it carries the hook and the skill together.

## Content split

| | |
|---|---|
| **SessionStart hook** | ~18 lines, always on: you are in a sandbox, `min search`/`min add`, the `--session` default, `min run`. |
| **`min-session` skill** | The fuller reference, loaded on demand — costs nothing in sessions that never touch packaging. |

The skill covers **the tool**, not packaging conventions — those stay in each repo's `AGENTS.md`, which Claude Code already discovers. Both defer to bare `min` output as the tiebreaker, since the shim ships with minimald rather than with this repo and will drift from anything hardcoded here.

## Why a separate package

- `claude-code` declares `license_spdx = "LicenseRef-Anthropic-Proprietary"` and `closed_source = true` for its whole tree. This content is Apache-2.0, and those attrs feed supply-chain tooling.
- The content tracks minimald's shim; `claude-code` tracks Anthropic's release cadence and is bumped by automation. Different clocks.

The usual objection — "now `claude-code` depends on the plugin" — is defused by making the wrapper conditional:

```bash
min_plugin=/usr/share/claude/plugins/minimal
if [ -z "${MINIMAL_CLAUDE_NO_PLUGIN:-}" ] && [ -d "$min_plugin" ]; then
  set -- "--plugin-dir=$min_plugin" "$@"
fi
```

`claude-code` keeps working standalone and gains no dependency. The edge points one way: plugin → `claude-code`. The hook also exits silently when `/usr/bin/min` is absent, so the plugin is inert outside a Minimal session.

## AGENTS.md fixes

Three of these were live bugs, verified against the shim (`crates/minimald/src/env_min_helper.sh`) and `crates/mctx/src/env.rs` in the `minimal` repo:

| Documented | Actual |
|---|---|
| `min patched-build <pkg>` | `min patched-pkg <pkg>` — the documented name is unroutable |
| `harnesses/<name>/harness.ncl`, `min check --harnesses` | `stacks/<name>/stack.ncl`, `--stacks`; `harness` survives as a stdlib alias |
| harness list of 17 | 21 stacks; was missing aeneas, cabal, ocaml, stack |

Plus two undocumented behaviours that reliably trip agents up: `min add` defaults to `--session` (ephemeral, persists nothing), and interactive tasks cannot be launched from inside a session — `min run shell` / `min run claude` are host-only.

## Verification

Done locally on macOS:

| Check | Result |
|---|---|
| `build.sh` assembles the plugin tree | 4 files, correct layout |
| Hook emits valid JSON | parsed with `json.load` |
| Hook guard with no `/usr/bin/min` | exit 0, silent |
| Plugin loads in real Claude Code | answered from the primer alone; skill registered as `minimal:min-session` |
| Wrapper, all three branches | absent → no flag; present → `--plugin-dir=…`; `MINIMAL_CLAUDE_NO_PLUGIN=1` → no flag |
| `ImportsCheck` rules | identifiers match folders, both imports used |
| Dotfile capture | `globset` with `literal_separator(true)` has no leading-dot exclusion, so `**` catches `.claude-plugin/` |

**Not run locally:** `min check` and `min patched-pkg`. The host `min` is session-management only and has no `check`; building `mip` from source fails on macOS because `cmd_run` is `#[cfg(target_os = "linux")]`-gated. Leaving those to the build server — `FmtCheck` (Nickel formatting) is the most likely thing to need a fixup.

## Summary by CodeRabbit

* **New Features**
  * Added a Minimal Claude Code plugin with sandbox guidance, session-start context, and in-session `min` command instructions.
  * Claude Code now automatically loads the Minimal plugin when available, with an environment option to disable it.
  * Added plugin support to shell and Claude tasks.

* **Documentation**
  * Updated CLI guidance, stack terminology, package workflows, and troubleshooting instructions to reflect current commands.


twitchyliquid64 (approved):
Zomg

PR #509, opened 2026-07-22
Add zvelte-check-bin 0.2.0

bryan-minimal:
A fast Svelte diagnostic tool written in Zig, from the `ampcode/homebrew-tap` Homebrew formula.

**First package produced by `pkgmgr import homebrew`** (pkgmgr-rs#567) — this PR is partly a test that the importer produces something a reviewer would accept by hand.

## What was verified rather than trusted

- **Both sha256 values were recomputed from the downloaded bytes**, not copied out of the formula. The tap is a cross-check, never the root of trust for the one value anchoring this package's integrity.
- **`strip_prefix` is absent because the archives are genuinely flat** — read off the real tarball (`tar -tzf` → one entry, `zvelte-check`). Homebrew auto-strips a single leading directory before `def install` and minimal does not, so this is unknowable from formula text and a guess would surface at build time as a missing file.
- **Provenance is derived from the artifact URLs**, which are literally `github.com/ampcode/zvelte-check/releases/…`, and cross-checked against the homepage. Never from the homepage alone — that conflation is how `chromium-bin` came to name `microsoft/playwright` for an artifact served out of Google's bucket. This buys real CPE identity (not a dark package) and free version checking via the GitHub tier.

## One thing I want to flag rather than bury

**This is prebuilt, and `zig` IS packaged.** pkgs CLAUDE.md prefers a source build when the toolchain exists. Prebuilt was chosen only because pkgmgr has no Zig `BuildFlavor` yet — *not* because the toolchain is missing, and I didn't want the build.ncl header to imply otherwise.

Mitigating: these are upstream's own release binaries, not a third-party redistribution, and we verify the hashes ourselves. A source-built `zvelte-check` can take the bare name later, which is why this carries `-bin`.

## No reviewer TODOs

Licence, provenance, outputs and deps all resolved mechanically. Requires no stdlib bump and touches nothing else in the repo.

## Summary by CodeRabbit

* **New Features**
  * Added the `zvelte-check` command-line tool as an installable package.
  * Provides prebuilt binaries for x86_64 and ARM64 Linux systems.
  * Installs the executable for direct use from the system command line.
  * Includes version 0.2.0 with architecture-specific downloads.
  * Verifies downloaded packages to improve installation reliability and integrity across supported systems.

### DIFF
--- a/packages/zvelte-check-bin/build.ncl
+++ b/packages/zvelte-check-bin/build.ncl
@@ -1,0 +1,51 @@
+let { Attrs, BuildSpec, Local, OutputBin, Source, .. } = import "minimal.ncl" in
+let { target, .. } = import "config.ncl" in
+let base = import "../base/build.ncl" in
+let glibc = import "../glibc/build.ncl" in
+
+# Imported from the Homebrew formula `ampcode/homebrew-tap/zvelte-check` by `pkgmgr import homebrew`.
+#
+# PREBUILT vendor binary — there is no source build. The pinned sha256
+# per architecture is the ONLY integrity anchor for this package.
+let version = "0.2.0" in
+let target_suffix = match { { arch = 'Amd64, .. } => "x86_64", { arch = 'Arm64, .. } => "aarch64" } target in
+{
+  name = "zvelte-check-bin",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    match {
+      { arch = 'Amd64, .. } =>
+        {
+          url = "https://github.com/ampcode/zvelte-check/releases/download/v%{version}/zvelte-check-linux-x86_64.tar.gz",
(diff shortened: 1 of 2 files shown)


jtnkminimal (approved):

jtnkminimal (approved):

twitchyliquid64 (approved):

packages/zvelte-check-bin/build.ncl
```diff
@@ -0,0 +1,49 @@
...
+
+  outputs = {
+    zvelte-check = { glob = "usr/bin/zvelte-check" } | OutputBin,
+  },
+
+  attrs =
+    {
+      upstream_version = version,
+      license_spdx = "MIT",
+      source_provenance = {
```
twitchyliquid64, line 43: If this is truly a binary fetch, drop source provenance and set `binary_from`. Also if its closed source you need `closed_source = true`.

PR #510, opened 2026-07-22
Add ampcode (Amp CLI) — DRAFT, proprietary + unscannable

bryan-minimal:
The Amp CLI, from the `ampcode/homebrew-tap` formula via `pkgmgr import homebrew`.

**Draft on purpose.** This package is honest about two things a *reviewer* must decide — not two things the tooling failed to work out.

## Both TODOs are resolved in the file, not left open

**CPE — searched, none exists.** NVD `keywordSearch=ampcode` returns **0 results**. There is no CPE identity for Amp, so the package is unscannable *by construction, not by omission*.

`sourcegraph` returns 462 CPEs — but every one is `sourcegraph:sourcegraph`, the code-search product. **A different product.** Attaching it would import another product's CVEs as false positives, so it's deliberately not done. That over-match is a failure mode we've spent real time cleaning up.

**Terms — read, and they're restrictive.** ampcode.com/terms grants a *"limited, revocable, non-exclusive, non-transferable, non-sublicensable"* licence. Redistribution is **not** permitted.

So the package fetches **direct from `static.ampcode.com` and is never mirrored to `gs://`** — the same shape `chromium-bin` uses against Google's and Playwright's CDNs. 210 of our 533 sources already fetch direct, so this isn't novel.

## No provenance, on purpose

A vendor CDN URL yields none, and fabricating a plausible one is exactly how `chromium-bin` came to name `microsoft/playwright` for an artifact served out of Google's bucket. Omission is honest; a wrong owner/repo is worse than nothing.

## The repo-wide bit — the reason this is separate

Requires **stdlib 0.0.18** for the `redistributable` attr class (minimal#861), hence the `minimum_version` bump. `Attrs` is a closed allowlist, so on an older stdlib this fails Nickel eval for the *entire repo* — I hit that, and it took out an unrelated sibling package's build. Buildbot is confirmed on 0.0.18.

⚠️ Nothing in minimal **reads** the attr yet (#861 is schema + docs), so today it records the constraint rather than enforcing it. Our binary cache is not yet mechanically prevented from carrying this artifact.

## Two operational facts before merging

- Artifacts are **~111 MB each, ~222 MB per version** across both arches, and the version is a unix timestamp that bumps on essentially every release — **it moved twice while this PR was being prepared**.
- `runtime_deps` is a discovery-mode guess of `[glibc]`. A 111 MB binary is likely a Node SEA or Bun-compiled blob; only a `--from-build` DT_NEEDED scan will tell the truth. Blocked on local build tooling (see below).

## Not yet built locally

`mip` (the package/build CLI) has no macOS artifact and doesn't compile on darwin — `crates/mip/src/main.rs` gates `mod cmd_run` on `target_os = "linux"` but leaves three call sites ungated. Investigating the correct build path separately; buildbot is the validation here.

### DIFF
--- a/minimal.toml
+++ b/minimal.toml
@@ -1,5 +1,5 @@
 [stdlib]
-minimum_version = "0.0.15"
+minimum_version = "0.0.18"
 
 [tasks.shell]
 interactive = true
--- a/packages/ampcode/build.ncl
+++ b/packages/ampcode/build.ncl
@@ -17,12 +17,14 @@ let glibc = import "../glibc/build.ncl" in
 #   same-vendor but different product — that imports another
 #   product's CVEs as false positives.
 # TODO: REVIEWER — formula declares `license :cannot_represent` (proprietary). Emitted
-#   as closed_source + redistributable=false — CONFIRM the
-#   vendor's terms before merging, and note nothing in minimal reads
-#   `redistributable` yet (#861 is schema only), so it does not
-#   currently prevent an upload.
+#   as closed_source + redistributable=false, which the prod
+#   buildbot ENFORCES (build-servers#180): the spec is excluded from
+#   cache upload, exempted from the cache audit, and skipped in the
+#   closure seal, so the artifact never enters the shared binary
+#   cache. CONFIRM the vendor's terms still match that posture
+#   before merging.
 #
-# ── TODO resolution (searched 2026-07-22) ───────────────────────────
+# ── TODO resolution (verified 2026-07-22) ───────────────────────────
 # CPE: NVD keywordSearch=ampcode returns 0 results — no CPE identity
 #   exists for Amp, so there is nothing to attach and this package is
 #   UNSCANNABLE by construction, not by omission. Note `sourcegraph`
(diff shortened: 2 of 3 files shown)


jtnkminimal (approved):
APPROVED FOR TESTING + DRAFT ONLY

bryan-minimal:
**Correction — `redistributable` IS enforced.** The PR body above says nothing in minimal reads the attr, so it records the constraint rather than applying one. That's wrong, and wrong in the direction that matters: it would lead a reviewer to conclude the proprietary binary could still reach the shared cache.

[build-servers#180](https://github.com/gominimal/build-servers/pull/180) merged 2026-07-21 and enforces it — specs with `attrs.redistributable = false` are **excluded from cache upload**, **exempt from the cache audit**, and **skipped in the closure seal** (unconditionally, so post-purge seals record absence rather than a seal-blocking hole). Consumers build locally via the planner's index-miss fallback.

minimal#861 *declares* the attr; the buildbot is what *acts* on it. I'd only read the stdlib side, which is how I got it backwards.

So two independent mechanisms now cover the licence constraint, and `fa64565` says so in the file:
1. the `Source` urls fetch **direct from static.ampcode.com** — the artifact is never mirrored to `gs://` in the first place;
2. `redistributable = false` keeps it out of the shared cache if it ever were.

Also confirmed: this package **cannot** be built locally on macOS at any shim version — `redistributable` needs stdlib 0.0.18 and the shim's stdlib is compiled into the binary. Buildbot is the validation here, which is why the `minimum_version` bump rides along.

PR #512, opened 2026-07-23
fix(libkrun): carry the packet-count vsock backpressure patch

norrietaylor:
Brings the **Linux** libkrun package in line with the vsock upload fix that shipped on the macOS source build ([gominimal/minimal#926](https://github.com/gominimal/minimal/pull/926), merged). Until now Linux carried the older fill-loop + signal-queue patches; this replaces both with the single packet-count backpressure patch.

## The bug (host-OS-independent)

`min activate` uploads over the guest vsock connection fail mid-transfer with `ENOBUFS` ([gominimal/minimal#869](https://github.com/gominimal/minimal/issues/869)). Since Linux 6.12.92, `virtio_transport_inc_rx_pkt()` resets the connection once `(skb_queue_len + 1) * SKB_TRUESIZE(0) > buf_alloc` — a **second, packet-count receive ceiling** that the vsock credit protocol (which meters bytes only) does not express. libkrun's `UnixProxy` fragments a bulk stream into sub-`SKB_TRUESIZE(0)` packets (576 B on arm64), exhausting the count ceiling with the byte window mostly unused. The fragmenting code and the guest kernel are identical across host OSes, so Linux VM-backed sessions hit this exactly as macOS did.

## The fix

Meter packet count the same way byte credit is metered. `recv_to_pkt()` tracks the guest's `skb_queue_len` exactly — the tail `rx_cnt` of each emitted packet, retired as `fwd_cnt` passes it — and takes the existing `WaitForCredit` path before emitting one packet too many, bounded by a conservative `buf_alloc / SKB_OVERHEAD`.

Unlike the previous fill-loop patch it does **not** change delivery granularity: `send()` stays 1:1 with `recv()`, nothing is delayed for batching. It is pure backpressure, invisible until the reader falls behind — the more defensible shape, and why #926 replaced the coalescing approach on the macOS side. It bundles the credit-request wakeup fix (`push_credit_req` must set `signal_queue`), which the count ceiling makes reachable during bulk transfer, so the two former patches collapse into one.

## Changes

- Remove `0001-vsock-signal-…` and `0002-vsock-fill-…`.
- Add `0001-vsock-bound-outstanding-packets.patch` (identical to the one in `gominimal/minimal` `vendor/libkrun/patches/`).
- `build.sh`: apply the single patch.
- `build.ncl`: one `Local` patch dep; comment rewritten.

## Verification

- Applies clean to the pinned `v1.19.4` tarball via `patch -Np1` (the exact mechanism `build.sh` uses) — dry-run touches only `src/devices/src/virtio/vsock/{unix.rs,tsi_stream.rs}`.
- The same patch, built and Developer-ID-signed through the macOS release pipeline, passed **100/100** uploads on the shipped release binary with the instrumented guest kernel logging **zero** receive-queue resets.

Note: I could not run a Linux `libkrun.so` build locally (macOS host). The patch application and source are verified; the Linux package build itself will be exercised by CI here.

Refs: gominimal/minimal#869, gominimal/minimal#926

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vsock connection reliability by preventing stalls during credit updates.
  * Added safeguards to prevent excessive outstanding packet queues and reduce backpressure issues.
  * Improved receive handling by combining available data into fuller packets, reducing fragmentation.
* **Performance**
  * Improved data transfer efficiency by filling receive buffers more effectively.
  * Enhanced flow control for smoother communication under higher traffic loads.


bryan-minimal (approved):
LGTM

PR #525, opened 2026-07-26
zola, tamarin-prover: fix reproducibility (parallel codegen order; compile-time clock)

bryan-minimal:
Both packages were flagged `unknown` by a build-twice audit of the 87 packages added since the last fleet run (83/86 reproducible, 96.5%). They looked identical in the verdict table and turned out to be completely different problems.

## zola — build-time entropy, not codegen order (corrected by build-twice)

The original analysis here blamed parallel-CGU emission order and proposed `-C codegen-units=1` + `-C symbol-mangling-version=v0`. Build-twice verification **falsified that**: with both flags applied, 17.4% of bytes still differed (size-preserving, twin-windows). Two further controls sharpened it — ripgrep builds reproducibly with *no* ordering flags at all (so CGU output is assembled deterministically by rustc), and symbol forensics on unstripped pairs showed identical symbol sets AND order with churn inside generated hash-table code (first address divergence at `minify_html_common`'s spec tables; size-flapping perfect-hash `asso_values`).

Mechanism: **build-time entropy** — HashMap-iteration randomness in build-time codegen, invisible to every rustc ordering flag. Fix (last commit): LD_PRELOAD a 12-line deterministic `getrandom`/`getentropy` interposer around `cargo build` only — the entropy analogue of `SOURCE_DATE_EPOCH`, exactly as minimal-repro's guide prescribes. The runtime binary is unaffected and the upstream profile (fat LTO, strip) stays untouched, so the fix costs zero runtime performance.

## tamarin-prover — not GHC, and not the toolchain

Worth stating plainly, since the GHC determinism work is recent: **that work is holding**. tamarin's Haskell codegen is bit-identical across builds. What differs is **26 bytes out of a 135 MB binary**:

```
Git revision: UNKNOWN, branch: UNKNOWN
Compiled at: 2026-07-25 03:47:54.541898723 UTC     <- build A
Compiled at: 2026-07-25 03:53:20.855637052 UTC     <- build B
```

tamarin's version banner embeds the wall-clock compile time through a TemplateHaskell splice that calls `getCurrentTime` **while compiling**. It asks the clock directly, so the sandbox's `SOURCE_DATE_EPOCH` never reaches it. (The git fields are already deterministic — there's no repo in the build tree, so both builds say `UNKNOWN`.)

This is the same family as perl's `cf_time` (#291): an application baking a timestamp, not a compiler being nondeterministic.

The patch rewrites the splice to a fixed instant derived from `SOURCE_DATE_EPOCH`, and **locates the source file by content** (`grep -rl 'Compiled at'`) rather than by path, so an upstream file move fails loudly here instead of silently reverting to a wall clock. It follows the verify-grep style the recipe already uses for its other source patches.

## Verification status — both halves verified

Forced-real build-twice via repro-lab on `min 0.5.1-rc1` (each build's cache slot evicted first; runs abort unless the builder proves a real build; every build byte-compared against build 1):

- **zola:** with the entropy shim — **4 real builds, all byte-identical**, upstream profile untouched.
- **tamarin-prover:** patch as shipped — **4 real builds, all byte-identical** (the builder's warm dep store means only tamarin's own modules rebuild, which is precisely the region the patch touches).

Disclosed caveats: verification used an inert `cmd`-field spec perturbation to defeat artifact hydration (changed build scripts do not change the spec hash on the current scheme — see the SpecHash migration discussion); and these are target-determinism builds (identical deps by construction), not clean-room. Full methodology, lever matrix, and forensics trail: repro-lab `docs/pr525-doublebuild-results.md`.

## Summary by CodeRabbit

* **Build Improvements**
  * Improved reproducibility for Tamarin Prover builds by ensuring version metadata is derived consistently from the build environment.
  * Improved reproducibility for Zola builds with deterministic compilation settings and stable generated outputs.
  * Added safeguards to detect unexpected build-time timestamps or unsupported compilation conditions.

### DIFF
--- a/packages/tamarin-prover/build.sh
+++ b/packages/tamarin-prover/build.sh
@@ -67,6 +67,36 @@ grep -v '^with-compiler:' stackage-lts-24.50.cabal.config > lts-pinned.config
 # so the field is reachable via the type. (pkgmgr-rs#528)
 sed -i 's/defaultTheoryLoadOptions, maudePath, TheoryLoadError/defaultTheoryLoadOptions, TheoryLoadOptions(maudePath), TheoryLoadError/' src/Main/REPL.hs
 
+# Reproducibility: tamarin's version banner embeds the WALL-CLOCK compile time
+#
+#     Git revision: UNKNOWN, branch: UNKNOWN
+#     Compiled at: 2026-07-25 03:47:54.541898723 UTC
+#
+# via a TemplateHaskell splice that calls getCurrentTime while COMPILING. It
+# asks the clock directly, so the sandbox's SOURCE_DATE_EPOCH never reaches it,
+# and two builds differ by exactly that string — measured: 26 bytes out of
+# 135 MB, with the Haskell codegen itself bit-identical. (The git fields are
+# already deterministic: no repo here, so both builds say UNKNOWN.)
+#
+# Rewrite the splice to a fixed instant derived from SOURCE_DATE_EPOCH. Located
+# by content rather than by path so an upstream file move fails loudly here
+# instead of silently reverting to a wall clock.
+STAMP="$(date -u -d "@${SOURCE_DATE_EPOCH:-0}" '+%Y-%m-%d %H:%M:%S UTC' 2>/dev/null \
+        || date -u -r "${SOURCE_DATE_EPOCH:-0}" '+%Y-%m-%d %H:%M:%S UTC')"
+stamp_file="$(grep -rl 'Compiled at' --include='*.hs' src lib 2>/dev/null | head -1)"
+if [ -z "$stamp_file" ]; then
+    echo "ERROR: no source file embeds 'Compiled at' — tamarin's version banner moved; revisit this patch." >&2
+    exit 1
+fi
+# Replace the runIO getCurrentTime splice with the pinned literal.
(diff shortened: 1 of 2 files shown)


twitchyliquid64 (approved):

bryan-minimal:
Build-twice results, as promised before leaving draft — both halves verified on `min 0.5.1-rc1` with forced-real builds (cache slot evicted before every build; the harness aborts unless the builder proves a real build happened; every build byte-compared against build 1):

- **zola**: the original `codegen-units=1` + `v0` fix was **falsified** (17.4% still differed). Root cause is build-time entropy in generated hash-table code (symbol-forensics attribution in the updated description). With the deterministic-entropy shim (`6ca9d7b`): **4 real builds, all byte-identical**, upstream profile untouched — zero runtime perf cost.
- **tamarin-prover**: the splice patch works exactly as shipped: **4 real builds, all byte-identical**.

The description's zola section has been rewritten to match the evidence (the ripgrep control disproves the CGU-completion-order explanation — worth knowing for the guide, which repro-lab will follow up on separately).

Caveats, disclosed: verification used an inert `cmd`-field spec perturbation to defeat artifact hydration (script-only changes don't move the spec hash on the current scheme), and these are target-determinism builds, not clean-room. Full lever matrix and forensics trail: repro-lab `docs/pr525-doublebuild-results.md`.

twitchyliquid64 (approved):

PR #545, opened 2026-07-30
ghidra: package the reverse-engineering suite 12.1.2

bryan-minimal:
The only credible open decompiler for native code, and the anchor of a reverse-engineering loadout. `analyzeHeadless` is the point — it makes Ghidra scriptable across a fleet of binaries rather than a GUI you drive by hand.

Apache-2.0 at the root; the tree aggregates ~20 further licences (GPL-2.0 WITH Classpath-exception, LGPL, BSD, MIT, Zlib, MPL-2.0, Python-2.0 …) — **all OSI/FSF-free, nothing proprietary**.

## Prebuilt release, not a source build

Ghidra's Gradle build runs `fetchDependencies.gradle` **before** compiling anything: 67 URLs across seven hosts, including binaries with no source (AXMLPrinter2.jar from the dead Google Code archive, per-platform Z3 zips), a postgresql tarball and PyPI wheels. It also requires JDK 25 to build; we ship 21.

The release zip is the same tradeoff the `jdk` package already makes, and it bundles **no JRE**, so Ghidra runs on our jdk rather than smuggling in a second Java.

## arm64 was not at parity, and now is

Upstream publishes natives for `linux_x86_64`, `mac_arm_64`, `mac_x86_64`, `win_x86_64` — and **not `linux_arm_64`**. On an arm64 host that is not a warning, it is a **silently degraded run**: `Application.getModuleOSFile()` has x86_64 fallbacks only for `WIN_ARM_64` (emulation) and `MAC_ARM_64` (Rosetta), so `decompile` is simply not found and `analyzeHeadless` exits **0** having produced no C.

Closed by building the five natives from C/C++ sources the release zip **already ships** — no second Source, no network, no Gradle. `ARCH_TYPE=` is the whole fix: the Makefile is `ifeq ($(ARCH),x86_64) -m64 else -m32` with an in-tree `TODO: need to revise to support arm64/aarch64`, so **`make ARCH=aarch64` would look right and do nothing** — every non-x86_64 value takes the `else`.

The 7-Zip gap was closable too, contrary to what an earlier revision of this PR asserted. Upstream's `-all-platforms` roll-up excludes ARM *by design*, but the per-platform artifact exists on Maven Central at the exact version Ghidra vendors.

## Three pre-existing bugs fixed along the way

- **`GPL/DemanglerGnu/os/` was never chmod'd** — verified in a built tree: both demanglers were mode **0644** while the decompiler natives were 0755. Ghidra `Runtime.exec()`s them by absolute path with no fallback, so **C++ demangling was silently broken**, on x86_64 too.
- **27 shell scripts had no exec bit** — every debugger launcher and all of `server/` — because `support/*` is a shallow glob. Now driven from the zip's own mode metadata, so it tracks upstream's intent. It correctly leaves `*setuputils.sh` alone: those are *sourced* libraries.
- **The win/mac prune missed the 7-Zip JNI blobs**, which use `Mac-x86_64`/`Windows-amd64` rather than Ghidra's `os/<platform>` convention.

## Tests, because the failures here are silent

Two, both mutation-tested (breaking each assertion makes `minimal check` fail):

**`headless_decompile`** — compiles a namespaced C++ binary, decompiles it, asserts real C comes back *and* that `Foo::bar` demangled. It is written against `decompileCompleted()` alone, because the alternatives lie: `isValid()` returns **true** for a decompile that produced nothing, and `isCancelled()` reports a *missing native* as a user cancellation. Nothing asserts on `analyzeHeadless`'s exit code — it has **zero `System.exit` calls** and exits 0 when the postScript throws.

**`sevenzip_native`** — proves the 7-Zip fix *works* rather than that a file exists. Its own output:

```
PLATFORM=Linux-arm64
ENTRY=SEVENZIP_PARITY_OK.txt
```

The platform assertion is load-bearing: if the jar swap regresses to shipping `all-platforms` alongside, the platform list stops being a single entry, `getPlatformBestMatch()` falls back to matching `os.arch` (`aarch64`) against `Linux-arm64`, and can never succeed.

## Review feedback applied

- **@twitchyliquid64: set `binary_from`** — correct, and ghidra was the only prebuilt package here not declaring it (chromium-bin, claude-code, gcloud, ampcode, android-sdk, cf, edgedelta all do). The comment records what it does *not* cover: the arm64 natives are compiled here, so the Java is upstream's binary and the arm64 decompiler is ours.
- **CodeRabbit: `zlib` duplicated in `build_deps` and `runtime_deps`** — also correct. Carried real risk since `sleigh` links `-lz` at build time, so I rebuilt rather than assumed: exit 0, 14/14, `sleigh` still links `libz.so`.

## Verified

```
minimal package --rebuild --no-fetch ghidra   exit 0
minimal check --packages ghidra               14/14 pass
natives                                       5/5 ELF aarch64, mode 0755
foreign-arch binaries in tree                 0
reproducibility                               byte-identical across rebuilds
installed size                                850M
```

## Known limits

- **850M installed** — by far the largest package in the repo. Worth a conscious call on cache/closure pressure.
- **Z3 symbolic execution** ships no `linux_arm_64`; fails *loudly*. Closing it means moving off the pinned Z3, whose own "arm64" asset contains x86-64 binaries — which is why the 7-Zip injection asserts `e_machine` on anything it adds.
- **The x86_64 path** takes upstream's natives via a `[ -f ]` guard and could not be exercised on this arm64 host.

Followed by a second PR adding gef, rizin and the `reveng` loadout, which needs this one to land first.

## Summary by CodeRabbit

* **New Features**
  * Added support for Ghidra 12.1.2.
  * Added native arm64 7-Zip archive handling.
  * Added launchers and platform-native libraries for supported environments.
  * Added headless native decompilation with GNU symbol demangling.
* **Tests**
  * Added validation for arm64 archive reading and native decompilation.
* **Documentation**
  * Added package metadata, including licensing and official source information.

### DIFF
--- a/packages/ghidra/build.ncl
+++ b/packages/ghidra/build.ncl
@@ -139,6 +139,83 @@ let version = "12.1.2" in
     # The C++ fixture is deliberately namespaced: `Foo::bar` in names.txt also
     # proves the GNU demangler natives ran, which nothing else here covers and
     # which shipped broken (mode 0644) for the life of this package.
+    # PROVE THE 7-ZIP PARITY FIX ACTUALLY WORKS, not merely that the .so is on
+    # disk with the right ELF arch — which is all the output-types checker and
+    # a `file` invocation can tell you.
+    #
+    # This is the one capability that was genuinely MISSING on arm64 (upstream
+    # excludes ARM from the -all-platforms roll-up by design), so "we injected
+    # a native" deserves more evidence than "a file exists". The failure it
+    # guards is silent twice over: SevenZipCustomInitializer catches Throwable
+    # and rewraps, initNativeLibraries() eats the result, and analyzeHeadless
+    # swallows the resulting IOException and exits 0 with an empty listing.
+    #
+    # Uses sevenzipjbinding's OWN documented entry point rather than Ghidra's
+    # wrapper, so it tests the jar + native pairing directly: if the platform
+    # list is not exactly one entry, getPlatformBestMatch() falls back to
+    # matching os.arch ("aarch64") against the artifact's "Linux-arm64" and can
+    # never succeed. That is precisely the mistake the build.sh comment warns
+    # about, and this catches it.
+    #
+    # The fixture is a real 187-byte 7z archive, embedded base64 because the
+    # sandbox ships GNU tar, which cannot WRITE 7z (bsdtar can; we do not
+    # package it). Decoding it needs nothing but coreutils.
+    sevenzip_native =
+      {
+        class = 'Standalone,
+        test_deps = [base, jdk],
+        cmds = [
+          [
+            "/bin/bash",
+            "-c",
+            m%"
+              set -eu
+              printf '%s' 'N3q8ryccAAOwgYeuFwAAAAAAAACEAAAAAAAAADYSLXkAOBhKq0FKVlNmTkeB6IoTIX///ploAAEEBgABCRcABwsBAAEjAwEBBV0AAIAADA0ACAoBDg+qJwAABQERLwBTAEUAVgBFAE4AWgBJAFAAXwBQAEEAUgBJAFQAWQBfAE8ASwAuAHQAeAB0AAAAFAoBAK3P1wCYIN0BEgoBAK3P1wCYIN0BEwoBAPnM1wCYIN0BFQYBACCApIEAAA==' | base64 -d > /build/fixture.7z
+              # Sanity: the fixture must itself be a 7z, or a green test would
+              # be proving nothing about 7z at all.
+              head -c 6 /build/fixture.7z | grep -q '7z' ||
+                { echo "fixture is not a 7z archive" >&2; exit 1; }
+
+              cat > /build/SevenZipCheck.java <<'JAVA'
+              import java.io.RandomAccessFile;
+              import net.sf.sevenzipjbinding.*;
+              import net.sf.sevenzipjbinding.impl.RandomAccessFileInStream;
+
+              public class SevenZipCheck {
+                  public static void main(String[] args) throws Exception {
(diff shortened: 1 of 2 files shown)


packages/ghidra/build.ncl
```diff
@@ -0,0 +1,250 @@
...
+let python = import "../python/build.ncl" in
+let toolchain = import "../toolchain/build.ncl" in
+let zlib = import "../zlib/build.ncl" in
+
+let version = "12.1.2" in
+{
+  name = "ghidra",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    # THE OFFICIAL PREBUILT RELEASE, not a source build — a deliberate choice.
```
twitchyliquid64, line 17: I think you need to set the `binary_from` attribute in this case

bryan-minimal: Good catch — added in 024fc72, and you were right that it was a real gap: ghidra was the only prebuilt package in the tree not declaring it (chromium-bin, claude-code, gcloud, ampcode, android-sdk, cf and edgedelta all do).

One nuance I documented alongside it, because ghidra is a mixed case:

```nickel
binary_from = "https://github.com/NationalSecurityAgency/ghidra/releases/download/Ghidra_%{version}_build/",
```

The Java tree is upstream's binary, but on arm64 the five platform natives (decompile, sleigh, both GNU demanglers, lzfse) are COMPILED HERE from the C/C++ sources inside that same release zip — upstream publishes no linux_arm_64 build at all. So `binary_from` on its own would slightly overstate what is prebuilt, and the comment says which half is which.

Rebuilt and rechecked after the change: 14/14.

twitchyliquid64 (approved):

PR #557, opened 2026-08-03
Add musl libc

lrishi:
## Summary
Adds build of muslc to enable 'alpine' like rootfs

## Changes

While this PR contains basic muslc build, I don't know if/see how we support compiling rest of the packages without glibc. If this PR causes issues with base libc functionality, I can also rename this to musl-gcc.

-

## Checklist

- [x] I've read [CONTRIBUTING.md](https://github.com/gominimal/pkgs/blob/main/CONTRIBUTING.md).
- [x] I've accepted the [ICLA](https://github.com/gominimal/pkgs/blob/main/legal/ICLA.md) (and CCLA if contributing on my employer's time). CLA Assistant will prompt me on this PR if I haven't already.
- [x] `min check` passes for the affected packages/harnesses.
- [x] `min patched-build <name>` succeeds for any package I added or modified.
- [x] For new packages: `source_provenance` points to the canonical upstream and the source builds from source (not a prebuilt release binary) where the required toolchain is available.
- [x] For version bumps: I've verified the new `sha256` against the upstream archive.

## Summary by CodeRabbit

* **New Features**
  * Added musl libc version 1.2.6 as a buildable package.
  * Provides binaries, libraries, headers, and the musl compiler wrapper.
  * Supports both dynamic and static compilation with validation tests.
  * Uses reproducible build settings and preserves runtime loader links.

### DIFF
--- a/packages/musl/build.ncl
+++ b/packages/musl/build.ncl
@@ -1,0 +1,67 @@
+let { BuildSpec, Local, OutputBin, OutputData, OutputLib, Source, Test, .. } = import "minimal.ncl" in
+
+let version = "1.2.6" in
+
+let base = import "../base/build.ncl" in
+let binutils = import "../binutils/build.ncl" in
+let make = import "../make/build.ncl" in
+let toolchain = import "../toolchain/build.ncl" in
+
+{
+  name = "musl",
+
+  build_deps = [
+    { file = "build.sh" } | Local,
+    {
+      url = "https://git.musl-libc.org/cgit/musl/snapshot/musl-%{version}.tar.gz",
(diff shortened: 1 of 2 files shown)


msample:
The contract for the build outputs from a package is that they don't overlap with those of any other package.

The tools in the Minimal Public Registry are based on glibc.  Coexistence with musl C is not something we have planned for.

What is your use case for having musl C? If it's producing an Alpine-like rootfs as a package output, see https://github.com/gominimal/pkgs/tree/main/packages/microvm-rootfs

twitchyliquid64:
(I'm curious to see the artifacts on the other side but agree with Mike that its not likely to gel well with the other packages)

lrishi:
Cool, I understand. I'll close this PR.
My use case was specifically that we ship a few apps which have to be statically compiled, and we have to use MIT licensed libs.

PR #575, opened 2026-08-06
fix(ghc): reduce build time with parallelism

twitchyliquid64:
- Use same flags for build and install, so the version that was built is the one called for in the install (previously we were doing a build and then throwing it away)
 - Set `-j $(nprocs)`


bryan-minimal (approved):
LGTM

PR #576, opened 2026-08-06
fix(bzip2): retire the prebuilt — build from source (24 to 23)

bryan-minimal:
## Summary

Retires the `bzip2` prebuilt. It builds from source now, taking foreign prebuilts in the base
soup from **24 to 23**.

## Why it was substituted, and why it needn't be

`bzip2` depended on the `base-bootstrap` bundle, and that bundle contains `bzip2` — so `bzip2`
was a member of its own transitive dependency set, and the planner substituted a prebuilt via
`replace_on_cycle`. The dependency was **bundle membership, not a build requirement**: bzip2's
`build.sh` untars a `.tar.gz`, which needs gzip, not bzip2.

Naming the tools it actually uses instead of the bundle breaks the cycle, so the breaker is
removed.

## Verification

Cold amd64 `base` build with this exact change:

| criterion | result |
|---|---|
| build succeeded | `rc=0` |
| bzip2 absent from hydrate list | `bzip2=absent` |
| really compiled (not a cached no-op) | `build_secs=447` |
| source actually fetched + built | `bzip2-1.0.8.tar.gz` |
| prebuilt count | **24 to 23** |

Gated *before* spending the build: a classifier self-test confirmed an untouched tree reports
schedulable, then the candidate tree (explicit deps **and** breaker removed) also reported
schedulable — confirming the explicit tool list does not reintroduce the cycle. Checked
separately on aarch64 and amd64, since `replace_on_cycle` is arch-gated.

## Scope

One file. No other package's recipe changes.

An exhaustive follow-up — deleting each `build_dep` of each of the remaining 23 in turn, with the
breaker removed — found **no** other package freed by a single edge cut. bzip2 was the last cheap
win of this shape; the remaining 23 need multi-edge or structural changes.

### DIFF
--- a/packages/bzip2/build.ncl
+++ b/packages/bzip2/build.ncl
@@ -1,6 +1,12 @@
 let { Attrs, BuildSpec, Local, OutputBin, OutputData, OutputLib, Source, .. } = import "minimal.ncl" in
 let { target, .. } = import "config.ncl" in
-let base-bootstrap = import "../base-bootstrap/build.ncl" in
+let bash-bootstrap = import "../bash-bootstrap/build.ncl" in
+let coreutils = import "../coreutils/build.ncl" in
+let sed = import "../sed/build.ncl" in
+let tar = import "../tar/build.ncl" in
+let gzip = import "../gzip/build.ncl" in
+let make = import "../make/build.ncl" in
+let diffutils = import "../diffutils/build.ncl" in
 let make = import "../make/build.ncl" in
 let toolchain = import "../toolchain/build.ncl" in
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
I could be wrong but i had in my memory that we needed bzip2 for tar, making it part of the circular base soup. idk tho

bryan-minimal:
> I could be wrong but i had in my memory that we needed bzip2 for tar, making it part of the circular base soup. idk tho

yeah, this was a premature change from my hex0 stuff, going to wait until I have all of them replaced with the hex0 rooted versions, I'll close this out.

PR #607, opened 2026-08-14
Add pi 0.84.2

0chroma:
Add the `pi` package, the Pi coding agent from pi.dev (npm `@earendil-works/pi-coding-agent`), version 0.84.2.

Pi is a minimal terminal coding harness. It is an npm package, so it is installed from the registry rather than built from the GitHub monorepo, whose build needs the `tsgo` native TypeScript compiler and a multi-package workspace. This matches the existing npm CLI packages here, with one difference: the tarball is pinned as a `Source` by sha256 (like `cf`), and the tarball ships an `npm-shrinkwrap.json` that locks the transitive dependency tree. The build is reproducible: two `--no-cache` builds produce identical content-addressed output.

| File | Purpose |
| --- | --- |
| `packages/pi/build.ncl` | Build spec: sha256-pinned registry tarball Source, `node` + `coreutils` runtime deps, `dns`/`internet` needs, `source_provenance` = `earendil-works/pi` |
| `packages/pi/build.sh` | Installs the pinned tarball into `usr/libexec/pi`, symlinks `usr/bin/pi` |

Verification:
- `mip check --packages pi` passes
- `mip package build pi` builds successfully; two `--no-cache` builds give byte-identical output

### DIFF
--- a/packages/pi/build.ncl
+++ b/packages/pi/build.ncl
@@ -1,0 +1,52 @@
+let { Attrs, BuildSpec, Local, Needs, OutputBin, OutputData, .. } = import "minimal.ncl" in
+let base = import "../base/build.ncl" in
+let coreutils = import "../coreutils/build.ncl" in
+let node = import "../node/build.ncl" in
+
+# pi is the Pi coding agent (pi.dev), published to npm as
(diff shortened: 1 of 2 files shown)


bryan-minimal (approved):
LGTM

PR #608, opened 2026-08-14
Add maki 0.4.7

0chroma:
Add a package for maki 0.4.7, a Rust TUI AI coding agent (github.com/tontinton/maki).

## Changes

| File | Purpose |
| --- | --- |
| `packages/maki/build.ncl` | BuildSpec for maki 0.4.7: GitHub source tarball, rust + toolchain + make + perl build deps, glibc + libgcc runtime deps, standalone `maki --version` test |
| `packages/maki/build.sh` | `cargo build --release --locked --bin maki` with Rust and C determinism flags (path remaps, prefix map) |

Notes:

- `make` and `perl` are required because isahc builds vendored OpenSSL from source (static-ssl). `toolchain` ships neither.
- Built-in Lua plugins are embedded at compile time via `include_dir!`, so the package ships only the binary.
- `needs = { dns, internet }` for cargo fetches; `Cargo.lock` is committed, so the build uses `--locked`.

## Verification

- `mip check --packages maki`: all 15 checks pass, including `missing runtime_deps` and the standalone test
- Two clean-room builds (`mip package build` and `mip package build --rebuild`) produce byte-identical output (same content-addressed hash; `diff -r` clean)
- `maki --version` prints `maki 0.4.7`

### DIFF
--- a/packages/maki/build.ncl
+++ b/packages/maki/build.ncl
@@ -1,0 +1,67 @@
+let { subsetOf, Attrs, BuildSpec, Local, Needs, OutputBin, Source, Test, .. } = import "minimal.ncl" in
+let base = import "../base/build.ncl" in
+let gcc = import "../gcc/build.ncl" in
+let glibc = import "../glibc/build.ncl" in
+# isahc (static-ssl) vendors and builds OpenSSL from source (perl Configure + make)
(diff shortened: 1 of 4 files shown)


bryan-minimal (approved):
LGTM

PR #650, opened 2026-08-26
fix: stamp Biome version at build time

### LINKED ISSUE #637: biome package reports Version: 0.0.0 — version not stamped at build
The `biome` package (2.5.8 in the catalog) builds a binary that reports `Version: 0.0.0` from `biome --version`. Consequence downstream: Biome emits an info-level "configuration schema version does not match the CLI version 0.0.0" on every run against any `biome.jsonc` that sets `$schema` (no schema value can match 0.0.0), which is permanent cosmetic noise for every consumer. Biome's build derives its version from build-time metadata (cargo workspace version / BIOME_VERSION env at build), so the spec likely needs to stamp it during the Rust build.

Found while migrating gominimal/webapp off the npm `@biomejs/biome` devDependency onto the catalog package.

https://claude.ai/code/session_01RRiSKQyQptTSdSGgD6CAwt

npetrovcode:
## Summary

Stamp the catalog version into the Biome Rust build so `biome --version` reports the packaged version instead of `0.0.0`. This prevents Biome from warning that a valid configuration schema version does not match the CLI.

## Related issues

Closes #637

## Changes

- Forward the package version to `build.sh` through `build_args`.
- Export it as `BIOME_VERSION` for Biome's compile-time version constants.
- Add a standalone regression test that requires the exact `Version: 2.5.8` output.

## Checklist

- [x] I've read [CONTRIBUTING.md](https://github.com/gominimal/pkgs/blob/main/CONTRIBUTING.md).
- [ ] I've accepted the [ICLA](https://github.com/gominimal/pkgs/blob/main/legal/ICLA.md) (and CCLA if contributing on my employer's time). CLA Assistant will prompt me on this PR if I haven't already.
- [ ] `min check` passes for the affected packages/harnesses.
- [ ] `min patched-build <name>` succeeds for any package I added or modified.
- [ ] For new packages: `source_provenance` points to the canonical upstream and the source builds from source (not a prebuilt release binary) where the required toolchain is available.
- [ ] For version bumps: I've verified the new `sha256` against the upstream archive.

## Notes for reviewers

Local checks completed: `sh -n packages/biome/build.sh` and `git diff --check`. The `min` helper, Cargo, and Nickel are not installed in this environment, so `min check` and a patched build could not be run locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Biome package version handling during builds.
  * Added validation to ensure the installed Biome command reports the exact expected version.

### DIFF
--- a/packages/biome/build.ncl
+++ b/packages/biome/build.ncl
@@ -1,4 +1,4 @@
-let { standaloneTest, subsetOf, BuildSpec, Local, OutputBin, Source, .. } = import "minimal.ncl" in
+let { standaloneTest, subsetOf, BuildSpec, Local, OutputBin, Source, Test, .. } = import "minimal.ncl" in
 let base = import "../base/build.ncl" in
 let make = import "../make/build.ncl" in
 let rust = import "../rust/build.ncl" in
@@ -31,6 +31,7 @@ let version = "2.5.8" in
   ],
 
   cmd = "./build.sh",
+  build_args = { include version },
 
   outputs = {
     bins = { glob = "usr/bin/biome" } | OutputBin,
@@ -52,5 +53,14 @@ let version = "2.5.8" in
 
   tests = {
     smoketest = standaloneTest "/usr/bin/biome --version",
+    version_is_exact =
+      {
+        class = 'Standalone,
+        test_deps = [base],
+        cmds = [
+          ["/bin/bash", "-c", "test \"$(/usr/bin/biome --version)\" = \"Version: %{version}\""],
+        ],
+      }
+        | Test,
   },
 } | BuildSpec
--- a/packages/biome/build.sh
+++ b/packages/biome/build.sh
@@ -3,6 +3,7 @@ set -ex
 
 export CC=gcc
 export LD=gcc
+export BIOME_VERSION="$MINIMAL_ARG_VERSION"
 export RUSTFLAGS="-C linker=gcc --remap-path-prefix=$(pwd)=/builddir --remap-path-prefix=$HOME/.cargo=/cargo"
 
 cargo build --release -p biome_cli


msample:
Hi @npetrovcode, thanks for the fix!  If you could accept the CLA (see above) we can integrate your change.

mitodrummer:
Thanks for this, @npetrovcode — the diagnosis was right and the fix was the correct one.

I'm closing this only because it has been blocked on the unsigned CLA since 2026-08-26 and we haven't been able to reach you. The work isn't being discarded: I've opened a replacement PR that keeps your approach, and #637 will close through it.

What carried over from here, unchanged in substance:

- `build_args = { include version }` to forward the packaged version into the build.
- `export BIOME_VERSION="$MINIMAL_ARG_VERSION"` in `build.sh` — this is the right knob. Both `crates/biome_cli/src/lib.rs` and `crates/biome_configuration/src/lib.rs` read it through `option_env!`, and `biome_cli`'s Cargo version is literally `0.0.0` in the source tree, so nothing else would have stamped it.
- The exact-match regression test on `--version`. Worth keeping exact rather than a substring, since a loose match would still pass against `0.0.0`.

What's different in the replacement:

- Rebased onto current main.
- Bumped 2.5.8 → 2.5.14, since six releases landed while this sat.
- Added a comment recording *why* `BIOME_VERSION` is load-bearing, so a future version bump doesn't drop it by accident.
- Ran what you couldn't in your environment: `min check` and a patched build.

If you'd still like the commit credit and are up for signing the CLA, say so here and I'm happy to close mine in favour of this one instead — the CLA is the only thing standing in the way. Sorry for the wait, and thanks for taking the time to report and fix it properly.

mitodrummer:
Closing in favour of #715, which carries this fix forward (see my note above). Thanks again, @npetrovcode.

PR #654, opened 2026-08-27
add `dotenvx`

jessie-minimal:
## Changes

-

## Checklist

- [ ] I've read [CONTRIBUTING.md](https://github.com/gominimal/pkgs/blob/main/CONTRIBUTING.md).
- [ ] I've accepted the [ICLA](https://github.com/gominimal/pkgs/blob/main/legal/ICLA.md) (and CCLA if contributing on my employer's time). CLA Assistant will prompt me on this PR if I haven't already.
- [ ] `min check` passes for the affected packages/harnesses.
- [ ] `min patched-build <name>` succeeds for any package I added or modified.
- [ ] For new packages: `source_provenance` points to the canonical upstream and the source builds from source (not a prebuilt release binary) where the required toolchain is available.
- [ ] For version bumps: I've verified the new `sha256` against the upstream archive.

## Summary by CodeRabbit

* **New Features**
  * Added dotenvx version 1.64.0 as an available build.
  * Supports Linux AMD64 and ARM64 architectures.
  * Includes the dotenvx command-line tool at `/usr/bin/dotenvx`.
  * Added version and environment-variable validation checks.

### DIFF
--- a/packages/dotenvx/build.ncl
+++ b/packages/dotenvx/build.ncl
@@ -1,0 +1,61 @@
+let { standaloneTest, Attrs, BuildSpec, Local, Needs, OutputBin, Test, .. } = import "minimal.ncl" in
+let { target, .. } = import "config.ncl" in
+let base = import "../base/build.ncl" in
+let curl = import "../curl/build.ncl" in
+let tar = import "../tar/build.ncl" in
+
+let version = "1.64.0" in
+let arch =
+  match {
+    { arch = 'Amd64, .. } => "x86_64",
+    { arch = 'Arm64, .. } => "aarch64",
+  } target
+in
+{
+  name = "dotenvx",
+  build_deps = [
+    { file = "build.sh" } | Local,
+    base,
+    curl,
+    tar,
+  ],
+  runtime_deps = [],
+  needs =
+    {
+      dns = {},
+      internet = {},
+    } | Needs,
+
+  cmd = "./build.sh",
+  build_args = {
+    include version,
+    include arch,
+  },
+
+  outputs = {
+    dotenvx = { glob = "usr/bin/dotenvx" } | OutputBin,
+  },
+
+  attrs =
+    {
+      upstream_version = version,
+      source_provenance = {
+        category = 'GithubRepo,
+        owner = "dotenvx",
+        repo = "dotenvx",
+      },
+    } | Attrs,
+
+  tests = {
+    smoketest = standaloneTest "/bin/dotenvx --version",
+    run =
+      {
+        class = 'Standalone,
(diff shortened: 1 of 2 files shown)


bryan-minimal (commented):
Thanks for this — the package structure is solid: per-arch mapping, typed `OutputBin`, provenance, and a real functional test (not just `--version`), which is more than a lot of packages ship with.

Two things I'd like to see changed before merge, and one note.

## 1. Blocking: the build fetches an unpinned installer over the network

```sh
curl -sfS https://dotenvx.sh/install.sh | sh -s -- ... --force
```

There's no integrity pin on the installer script or on the binary it downloads, so:

- **Not reproducible** — two builds of this commit can produce different bytes if upstream changes the installer or re-cuts the release. The repo requires byte-identical rebuilds.
- **No supply-chain verification** — we execute whatever `dotenvx.sh` serves at build time. That matters more than usual here: dotenvx manages secrets, so it's the last package where we'd want an unverified fetch.
- It's why the spec needs `needs = { dns, internet }`, which almost nothing else does.

**Upstream already publishes everything needed to avoid this** — release tarballs plus a `checksums.txt`. Official sha256s for v1.64.0:

```
linux-x86_64   666caa10d4c9fd97535982e989ad304bf916b9ab0611536e2f6adbc7ffc22c18
linux-aarch64  a16680bbe15d9183bff02b9ad6e71f88cf0e4e2fced04d9fd54340ebbb518de8
```

So this can become per-arch `Source` entries with pinned `sha256` (same shape as `jdk`), and `build.sh` reduces to extract-and-install. That also lets the `needs` block go away entirely.

I checked the tarball layout: it's flat (top-level `./dotenvx`), so no `strip_prefix` is required.

## 2. Missing `license_spdx`

`attrs` has `upstream_version` and `source_provenance` but no license. Upstream is **BSD-3-Clause**:

```nickel
license_spdx = "BSD-3-Clause",
```

We track license coverage across the fleet, so a new package landing without one puts it straight into the backfill queue.

## 3. Note: `runtime_deps = []` is correct — don't "fix" it

Flagging this because it looks like an omission and isn't. The binary is statically linked:

```
dotenvx: ELF 64-bit LSB pie executable, ARM aarch64, static-pie linked, stripped
```

So there's genuinely no glibc dependency. Worth a short comment in the spec so the next reviewer doesn't add one.

## Minor

The two tests disagree on how they invoke the binary — `smoketest` uses `/bin/dotenvx` while `run` uses bare `dotenvx` (PATH-dependent). Worth making them consistent.

jessie-minimal:
dotenvx 1.64.0 was added by #655

PR #659, opened 2026-08-28
Redirect pnpm, bun, deno, and gem global installs to ~/.local per the inbox#584 decision

agent-137:
Implements the decision recorded in gominimal/inbox#584 for four ecosystems — follow-up to gominimal/inbox#559, whose npm half is #657: global installs redirect to `~/.local` via each tool's narrowest honored knob, delivered at the package level. Every knob and failure mode below was verified in a live session before being written down.

## What breaks today

| Tool | Failure in a session |
|---|---|
| pnpm | refuses: `global bin directory ".../pnpm/bin" is not in PATH` (PATH is fixed; `pnpm setup` has nothing to edit) |
| bun | `bun add -g` succeeds into a cache-derived dir whose `bin` is off PATH — binaries never resolve |
| deno | `deno install -g` targets `~/.deno/bin`, off PATH; deno's own hint ("Add ... to PATH") is impossible to follow |
| gem | auto-falls back to user install (good), but binstubs land in `~/.local/share/gem/ruby/<v>/bin`, off PATH |

## The fixes

- **pnpm / bun / deno** — `/usr/bin` wrappers defaulting the one env var each tool honors (`PNPM_HOME`, `BUN_INSTALL`, `DENO_INSTALL_ROOT` → `$HOME/.local`), exec'ing the real entry point from `/usr/libexec/<pkg>/`. Each tool derives its global bin dir as `<var>/bin`, so bins land in `~/.local/bin`, already on the session PATH; a caller's own value always wins (`: "${VAR:=$HOME/.local}"`).
- **gem** — no wrapper needed: RubyGems' system config path is `/usr/etc/gemrc`, *inside the package*, so the ruby package ships one with `gem: --bindir ~/.local/bin` (the `~` expands; the flag is harmless on non-install commands; uninstall removes the stub symmetrically; `GEMRC`/`~/.gemrc`/CLI still override).
- Every package gains a regression test pinning its default; bun/deno `libexec` outputs carry `allow_executable = true` (a Data output matching a binary fails `min check` without it), and their standalone smoketests target the raw ELFs since `standaloneTest` runs without a shell.
- bun's `bunx` stays an argv0 symlink beside the real binary, preserving name-based multiplexing (verified live through an exec'd wrapper).

Not included, deliberately: **go** (finds `GOROOT` relative to its own executable, and half the catalog builds with it — needs its own PR and verification) and **cargo** (blocked on whether `rust` sessions should link at all — `cc`/`ld` are absent today). Both tracked in gominimal/inbox#584.

An earlier revision did the pnpm half as a stack-level `build_env_vars` entry; reverted in favor of package delivery (covers stack-less consumers, `$HOME` not a literal `/home`). History carries both steps; squash-merge collapses them.

## Verification (live session)

- pnpm: exact emitted wrapper exercised — `bin -g` → `/home/.local/bin`; `add -g`/`rm -g` round-trip; preset `PNPM_HOME` untouched; `pnpx` works.
- bun: `add -g nanoid` → `/home/.local/bin/nanoid` runs; default confirmed off-PATH; bunx argv0 multiplexing through the wrapper confirmed.
- deno: `install -g` → `/home/.local/bin`, runs; env-probe test command confirmed working.
- gem: install → stub in `~/.local/bin`, runs; `gem list`/`gem env` unaffected by the gemrc; uninstall removes the stub.
- shellcheck: all added blocks clean; pre-existing findings untouched.

https://claude.ai/code/session_01JGTXbdK7hgDupktNZEkALY


agent-137:
CI failure post-mortem (both failures were mine, both fixed in d5142f9):

1. **minimal-check / missing runtime_deps (pnpm, bun, deno)** — the new `#!/bin/sh` wrappers need a runtime provider for `bin/sh`; the checker caught that none was declared. `bash` (which ships the `sh` symlink) is now in each package's `runtime_deps`.
2. **minimal-check / standalone tests (ruby-lsp)** — the ruby package's new system gemrc leaked into ruby-lsp's vendored build: its `gem install` picked up the global `--bindir ~/.local/bin` default, wrote the binstub into the build sandbox home, and `/usr/bin/ruby-lsp` then exec'd a missing path. Fixed with an explicit `--bindir "$GEMS/bin"` (CLI beats gemrc). ruby-lsp is the only build-time `gem install` in the catalog, so no other package needs the same treatment.

@bryan thanks for the re-trigger — my fix is rebased on top of your no-op, so the new run includes both. The 2h bun-build hang looks unrelated to the diff content (the bun build.sh changes only touch the install step after `bun run build:release`); if it hangs again on this run it's worth a look independent of this PR.

bryan-minimal:
Heads-up: this now conflicts with main (`CONFLICTING` / `DIRTY`) on all four files it touches, after three PRs landed today. The conflicts are mechanical, but **one of them has a wrong resolution that would re-open a two-day outage**, so it seemed worth writing down rather than letting it get resolved by reflex.

Also worth saying: the last CI run here completed **2026-08-28**, so its red is stale and predates everything below.

## The one that matters: don't restore `node-lts` in pnpm

`packages/pnpm/build.ncl` on main is now:

```nickel
runtime_deps = [
  subsetOf node ["node"],   # NOT node-lts, and NOT the whole package
  coreutils,
],
```

This branch still has `node-lts`. Merging that back would undo #671.

**Why it matters** (pkgs#665/#670/#671): `next` and `agent-browser` both declare `node` themselves, so when pnpm also carried `node-lts`, two npm installs landed in one sandbox at the same paths. The sandbox hardlinks closure members **first-writer-wins per file**, so `/usr/lib/node_modules/npm/` came out as a blend of both versions — `minipass-flush/index.js` from npm 11.17.0 (which imports minipass by its 7.x *named* export) beside a nested `minipass@3.3.6` from npm 11.11.1 (default export only). `class Flush extends undefined` throws, and **every `npm install` dies**. Both packages were individually fine; only the merge was broken. It blocked `next` for two days and took four PRs down with it.

The `subsetOf` form means pnpm contributes only the `node` binary and **can never contribute an npm tree at all**, so the failure is structurally impossible rather than just currently-absent.

**Suggested resolution:** keep main's `subsetOf node ["node"]`, then re-apply your `bash` addition and the `global_bin_dir` test on top. Your changes and #671 are orthogonal — nothing here depends on it being `node-lts` specifically.

## The other three are additive

* `packages/pnpm/build.sh` — your `PNPM_HOME` wrapper; main is unchanged here apart from #671's comment block.
* `packages/bun/build.ncl` — main added three `OutputData` globs for completions (`bash_completion`, `fish_completion`, `zsh_completion`) and `OutputData` to the `minimal.ncl` destructure. Your `libexec` glob and `OutputData` import sit alongside them; keep both sets.
* `packages/bun/build.sh` — main appends completion installs from the committed `completions/bun.{bash,zsh,fish}`. Those read from the extracted source, so they are independent of where the binary ends up.

## Two things I noticed, offered as questions not objections

**Your `standaloneTest` fix is a real find** — *"standaloneTest runs with no shell in the sandbox, so a `#!/bin/sh` wrapper would ENOENT there."* That is the kind of thing that would otherwise show up as a baffling test failure much later.

**Adding `bash` to `runtime_deps` of both pnpm and bun widens what they impose on every consumer.** That is the same axis #671 just narrowed pnpm along, so it is worth a deliberate look rather than an accident. Concretely: `bash` and `bash-bootstrap` both ship `/usr/bin/bash` and `/usr/bin/sh`, and the two binaries genuinely differ (1,222,160 vs 1,133,600 bytes — `--with-installed-readline` vs `--disable-readline`). A fleet audit found only `e2fsprogs` and `mtools` currently hold both, so this is not firing today; the question is whether pushing `bash` into every pnpm/bun consumer's closure creates new pairs. If the wrapper only needs `/bin/sh`, a narrower dep would keep the surface small.

No action needed from me — flagging because the pnpm resolution is easy to get wrong and expensive when it is.

mitodrummer:
Rebased onto main (`9abf7d0c`), now `MERGEABLE`. Thanks for writing the pnpm one down — it is exactly the resolution reflex would have gotten wrong.

## Conflicts, as resolved

* **`packages/pnpm/build.ncl`** — kept main's `subsetOf node ["node"]` **and** the comment block explaining why, verbatim. `node-lts` is gone from this branch; #671 stands. My two additions sit on top: `bash` in `runtime_deps` and the `global_bin_dir` test. Nothing here depended on it being `node-lts` — you were right that they are orthogonal.
* **`packages/bun/build.ncl`** — both sets of outputs kept: your three completion `OutputData` globs plus my `libexec` glob (with `allow_executable = true`). `OutputData` was already in the destructure from your side, so the import line needed no change.
* **`packages/bun/build.sh`** — both blocks kept. The libexec install + wrapper emission runs first, then the completions install from `completions/bun.{bash,zsh,fish}`. They are independent: the completion files are read from the extracted source, not from wherever the binary landed.
* **`packages/pnpm/build.sh`** — additive, no interaction with #671's comment block.

`min check` passes on all nine touched files. The output-dependent checkers (`enumerate bins`, `output types valid`, `missing runtime_deps`, `standalone tests`) report `Skip` locally because nothing is built in this session — CI runs them for real.

## On `bash` in `runtime_deps`

I went looking for the narrower dep you suggested, and it does not exist:

* **`subsetOf bash ["sh"]` would ship a dangling symlink.** `packages/bash/build.sh:25` is `ln -sf bash $OUTPUT_DIR/usr/bin/sh` — `sh` *is* a relative symlink to `bash`, so a subset carrying the `sh` output without the `bash` output resolves to nothing.
* **`subsetOf bash ["sh", "bash"]` would not narrow the axis you care about.** It drops `mans`, `includes` and `locales` — real bytes, but not the ones in play. The `bash`/`bash-bootstrap` pairing you found is a collision on exactly `/usr/bin/bash` and `/usr/bin/sh`, and any subset that gives a working `/bin/sh` ships both of those paths by construction.
* **Nothing else in the catalog provides `/bin/sh`.** No `dash`, `busybox`, `toybox` or `ash` package exists; `bash` and `bash-bootstrap` are the only two providers, and `bash-bootstrap` is a toolchain-bootstrap package, not something to push into consumer closures.

So the choice is `bash` or no wrapper. For bun and deno there is no no-wrapper option — the thing that has to set the env var runs *before* an ELF starts, which means a shell. `bash` in `runtime_deps` is also the established idiom rather than a novel imposition: 28 packages already do it, including `python`, and `subsetOf bash` appears zero times in the repo.

Where I think your concern actually lands is the `bash`/`bash-bootstrap` pairing itself — it is a property of those two packages both owning `/usr/bin/{bash,sh}`, and it will keep being reachable through any of those 28 packages regardless of what this PR does. Happy to open a separate issue to track making that pair structurally impossible (an `sh`-only provider, or having `bash-bootstrap` stop claiming those paths once the real `bash` exists) if you think it is worth one — but it seems wrong to make this PR the place that gets solved.

## CI

The red on this PR predates the rebase (last run 2026-08-28, as you noted). A fresh run should pick up the current tip.

norrietaylor (approved):

PR #667, opened 2026-08-31
rust + go: bootstrap both language toolchains from the hex0-rooted, CS-attested chains

bryan-minimal:
# rust + go: bootstrap both toolchains from source, starting at hex0

`packages/rust` and `packages/go` currently start from a downloaded binary compiler (rust's stage0 snapshot from static.rust-lang.org, go's bootstrap toolchain from go.dev). This PR replaces both with chains of in-tree packages that begin at the 229-byte `hex0` seed and end at the compilers `rust` and `go` build with today. No recipe in the chain touches the network at build time.

## What is in the PR

**38 new packages, 4 modified.** They form four ladders plus two flips.

1. **stage0 ladder** (`stage0-*`, amd64): `hex0 → hex1 → kaem → mes → mescc → tcc 0.9.26 → tcc 0.9.27 → musl 1.1.24 → binutils 2.30 → gcc 4.0.4 → gcc 4.7.4 → gcc 10.4.0 → gcc 15.2.0`, with `gmp-mpfr-mpc`, `linux-headers`, `binutils-2.41`, `musl-1.2.5` and three tcc/musl relink steps (`-s1/-s3/-s4`) as the intermediate rungs. Each package builds only with what the previous rung installed. The early rungs are `kaem` scripts (`*.kaem`), the rest are `build.sh`.
2. **glibc-hosted toolchain**: `glibc-bedrock-2.42`, `binutils-2.46-glibc`, `gcc-15.2.0-glibc` (C and C++). Built by the top of the stage0 ladder, installed under versioned prefixes (`/usr/lib/glibc-bedrock-2.42`, …) so they never collide with the distro toolchain. Every binary they produce is linked with an explicit dynamic linker and rpath into that prefix.
3. **rust ladder**: `mrustc` (pinned upstream commit plus a patch series for aarch64 and shared-library builds, each patch documented in its header) builds `mrustc-rustc-1.90.0` (rustc 1.90.0 + cargo + std). From there `rustc-1.91.1 → 1.92.0 → 1.93.1 → 1.94.1 → 1.95.0 → 1.96.0`, each rung a standard `./x.py build` whose stage0 is the previous rung's install prefix. Each rung installs to `/usr/lib/rustc-<version>` and ships a functional gate (compile *and run* a program that exercises generics, trait objects, unwinding, TLS, proc-macros and cargo; the pass value is computed, never a literal).
4. **go ladder**: `go-1.4` (C-only root, built with the glibc-hosted gcc) `→ go-1.17.13 → go-1.20.14 → go-1.22.6 → go-1.24.9`, each using the previous rung as `GOROOT_BOOTSTRAP`.
5. **`packages/rust`**: stage0 is now the `rustc-1.96.0` rung (`bootstrap.toml [build] rustc/cargo` point at its prefix, so x.py skips the snapshot download); the seed tarball Sources and the download fallback are gone, the stage0 is asserted by `--version` before the build starts, and cargo runs with `CARGO_NET_OFFLINE=true`. Runtime deps and outputs are unchanged.
6. **`packages/go`**: bootstraps from `go-1.24.9` on amd64 and from the sealed arm ladder artifact on arm64; the go.dev bindist Source is gone and a missing bootstrap is fatal. Version moves 1.26.6 → 1.27.0, which needed two small source bridges in `otel-collector` and `trivy` for go 1.27 API removals.

`hex0` is the single binary trust root: a sha-pinned Source (amd64 seed and the sealed aarch64 seed).

## How to review it efficiently

- **Clones.** The rustc rungs, the go rungs and the two musl variants are byte-identical apart from version strings and stage0 pointers. Read one of each (`rustc-1.93.1`, `go-1.20.14`, `stage0-musl-1.1.24`) and `diff -r` it against its siblings; the diffs are a few lines each.
- **The two mechanisms that matter** are in `packages/rustc-*/build.sh` (and, for stage0 pinning, `packages/rust/build.sh`): (a) stage0 pinning via `bootstrap.toml`, asserted at start by `--version` and `--print sysroot` against the expected prefix; (b) the rungs' offline harness: network tools on PATH replaced by stubs that exit non-zero, `CARGO_NET_OFFLINE=true`, and empty `.git` markers inside vendored `*-sys` crates so their `build.rs` never runs `git submodule update`.
- **Outputs.** Every chain package exports only its own versioned prefix (one recursive `OutputData` glob). Ladder packages that legitimately ship ELF files as data (loaders, crt objects, musl's empty `crti.o`) use `allow_executable` / `allow_data` explicitly.
- **Architecture scope.** The stage0 and rung packages are amd64-only and export empty collections on arm64. `rust` and `go` on arm64 consume the sealed arm-ladder artifacts as pinned Sources; the arm64 rust rungs above 1.90 are not built from source yet and are the next tranche.

## Build cost

Ladder rungs take minutes each; `mrustc-rustc-1.90.0` about an hour; each rustc rung 30 to 60 minutes; `rust` and `go` about an hour each. Cold, the whole chain is roughly a day of sequential build time; with a warm artifact cache only changed links rebuild. Byte-seal `.answers` files are present but unpinned; pinning is a data-only follow-up once each rung has two identical cold builds.

## Follow-ups (not in this PR)

- `attrs` metadata (`license_spdx`, `upstream_version`) for the stage0 and rung packages.
- Express the tcc `-s1/-s3/-s4` inputs as pristine source plus patch files instead of prepared tarballs.
- arm64 rust rungs above 1.90 from source; pin the byte seals.


bryan-minimal:
Drive-by from a fleet-wide audit of sandbox path collisions — **you already know about this one**, so this is measurement rather than discovery. Three of your own build scripts document it:

```
stage0-tcc-0.9.27-musl-s1/build.sh:32   "…rung's headers first-writer-wins per sandbox; a glibc draw
                                         made tcc die in bits/*.h with the…"
stage0-tcc-0.9.27-musl-s3/build.sh:31   "When glibc's stdio.h won, TM1 died in glibc's bits/*.h chain…"
stage0-gcc-4.0.4/build.sh:179           "(glibc's /usr/include/bits/errno.h -> absent linux/errno.h).
                                         CPATH/LIBRARY_PATH prepend R4b's clean…"
```

What you may not have is how large this is fleet-wide. Measured on `res-server-amd64` over the retained journal:

* **30.6% of every hardlink collision in the fleet** comes from the stage0/bedrock family — second only to the node/node-lts npm blend (66.9%), which is a known two-day outage now fixed in #671.
* **glibc lost 3,207 hardlinks across 32 bootstrap sandboxes.**
* Of the colliding `usr/include` paths, **~79% differ in content** — glibc 2.44 vs musl is 188 of 189 differing; glibc 2.44 vs glibc-bedrock 2.42 is 931 of 1,408.

## Why the per-rung workarounds may be load-bearing in a fragile way

The composition is first-writer-wins **per file**, and the order comes from a `BTreeSet<PathBuf>` keyed on **store path — i.e. content hash** (`crates/op/src/specs.rs`: *"hardlinks them first-writer-wins, so iteration order must be deterministic"*; `crates/common/src/lib.rs` `hardlink_dir_contents` → `AlreadyExists => warn!`).

Deterministic for a given closure, but the winner **flips when either package is rebuilt**, because the hash changes. So "a glibc draw made tcc die" isn't a one-off you fixed — it's a coin whose current face depends on hash ordering. That's also why the `-nostdinc` / `CPATH` prepends work: they make the rung independent of who won, which is the right instinct.

By my count on this branch, **13 of 15 stage0 rungs** carry a `-nostdinc` or `CPATH` mitigation. Worth confirming the other two are genuinely insulated rather than currently-lucky.

## The structural half

Seven packages on this branch export the **unrestricted** `usr/include/**`:

```
glibc-bedrock-2.42        stage0-musl-1.1.24
stage0-binutils-2.30      stage0-musl-1.1.24-cc2
stage0-binutils-2.41      stage0-musl-1.2.5
stage0-mes
```

Your gcc and linux-headers rungs already show the better pattern — `usr/include/c++/**` and `usr/include/*/**` — which cannot collide with libc top-level headers at all. Narrowing the musl/mes globs the same way would make the isolation **structural** rather than positional: no draw to lose, instead of winning the draw via env vars.

I have not built this branch, so treat the glob suggestion as a question rather than a finding — you know the chain's constraints far better than I do, and there may be a reason the sysroot has to be exported whole. The measurements above are firsthand.

Context: `minimal` currently resolves these collisions with a `warn!` that names two filesystem paths and no package, which is why 100k+ collisions went untriaged. A staged fix (attribute → count → error on divergent `OutputBin`/`OutputLib`) is being written up separately; flagging here only because this PR is still a draft and the glob change is much cheaper before it lands than after.

bryan-minimal:
Verified your measurements against the tree — the numbers hold up, and the "currently-lucky" framing caught something real. Ground truth on both questions:

**The 2-of-15 without `-nostdinc`/`CPATH`:**
- `stage0-hex1` — immune by construction: it's a pure hex0 byte-transform (no C compiler, no headers in the build at all).
- `stage0-tcc-0.9.27-musl-s1` — **you're right to flag it, and my own mitigation is weaker than I claimed.** `mes.kaem` installs the mes libc headers at *top-level* `/usr/include` (only 9 files live under `/usr/include/mes/`), so the mes-first `-I` ordering added in 96026d35 insulates just those 9 names; `stdio.h` and friends still ride the draw. s1 passes today because the current closure's BTreeSet order gives mes the win — exactly your "coin whose current face depends on hash ordering." A real fix needs an uncontended mes header copy (tedious in kaem — no loops, explicit per-file copies) or the structural change below.

**The 7 unrestricted `usr/include/**` exports — consumer-verified, one at a time:**

| pkg | top-level readers remaining | verdict |
|---|---|---|
| `stage0-musl-1.1.24` | none (s3 moved to its §E versioned sysroot in 96026d35; cc2 builds `-nostdinc -Iinclude` from its own tree; s4 uses R4b) | **narrowable now** |
| `stage0-musl-1.1.24-cc2` | exactly one: `stage0-gcc-4.0.4:161` `LIBGCC2_INCLUDES=-I/usr/include` | repoint that at `/usr/lib/musl-bedrock/include`, then **narrowable** |
| `stage0-musl-1.2.5` | gcc-4.7.4/10.4.0/15.2.0, gmp-mpfr-mpc, linux-headers, binutils-2.41, B4 — all carry wrappers; need per-wrapper verification they use the versioned `musl-bedrock-1.2.5` path before narrowing | likely narrowable, unverified |
| `stage0-binutils-2.30/2.41` | headers are `bfd.h`/`ansidecl.h`-class (non-libc names, can't collide with the anchor) | narrow to an explicit list, cosmetic |
| `stage0-mes` | s1 (load-bearing, see above) | the hard one — needs the kaem sysroot copy first |
| `glibc-bedrock-2.42` | it IS the amd64 toolchain anchor — winning `/usr/include` in ordinary sandboxes is its job | stays wide |

One trap for whoever implements: **output globs are not part of the spec hash**, so narrowing a glob alone leaves the cached artifact shipping the old wide capture — each narrowed pkg needs a build.sh touch to force re-capture.

Holding the actual edits until the rest of the incoming review notes land so it's one batched commit; the s1 correction + musl-1.1.24 narrowing + the gcc-4.0.4 repoint are queued first.

0chroma (commented):
## Review: rust + go hex0-rooted bootstrap chains

Reviewed the full branch. CI is green (`minimal-check` and a full `minimal build` pass); note CodeRabbit skipped this PR entirely — 236 files exceed its 150-file limit — so this needs human eyes.

### Verified correct

- **Ladder consistency.** Every `GOROOT_BOOTSTRAP` / stage0 prefix resolves to a package present in that file's `build_deps`; no dependency cycles. Min-bootstrap versions line up: go1.27 needs ≥1.24.6, the 1.24.9 rung's own `minBootstrap` is exactly its predecessor 1.22.6, earlier rungs still accept the `go1.4` default.
- **Hashes.** Source shas recomputed against upstream for go 1.17.13/1.20.14/1.22.6/1.24.9/1.27.0, glibc-2.42, binutils 2.30/2.41/2.46, gcc 4.0.4/4.7.4/10.4.0/15.2.0, musl, gmp/mpfr/mpc, mes, tcc, linux-6.12.43 — all match. Arch prebuilt shas are pinned and match their URL path segments.
- **Stage0 pinning.** `--version` / `--print sysroot` / `src/version`+`src/stage0` assertions abort on mismatch; stub PATH dirs are genuinely prepended with the curl/git shadowing asserted; `CARGO_NET_OFFLINE=true`; the `vendor.rs` sed patch fails closed (grep before and after, `exit 1`).
- **Gates.** Pass values are computed from guarded sums, never literals; each gated command is `rc`-captured so `set -e` can't short-circuit.
- **go-1.27 bridges.** trivy 0.74.0 has exactly the two `json.SkipFunc` sites the rewrite targets and `errors.ErrUnsupported` is the documented replacement; otel 0.159.0's manifest has a `replaces:` block and no swiss mention, and the pinned swiss pseudo-version carries the right build guard.

### Findings

1. **Dead arm64 prebuilt branches** — `glibc-bedrock-2.42/build.ncl`, `binutils-2.46-glibc/build.ncl`, `gcc-15.2.0-glibc/build.ncl`: the inner `match` supplies an `'Arm64` prebuilt `Source`, but the outer `match` collapses arm64 to an empty `BuildSpec`. Those Sources are never fetched and the `build.sh` aarch64 blocks are unreachable. Harmless functionally (arm64 doesn't consume these), but the comments claiming "on aarch64 the prebuilt is extracted instead" are false.
2. **`stage0-tcc-0.9.27-musl-s1/build.sh:38`** — on extract/`cd` failure it copies `/usr/bin/tcc-0.9.26` to `$OUTPUT_DIR/usr/bin/tcc-musl` and **exits 0**, so a failure mode installs a mislabeled binary that would be cached. Every other path in this PR fails shut; this one should too.
3. **s1/s3 are retry-driven, not gated** — they run under `set +e` with an env-layout perturbation loop (4–6 tries) and ship no `stage0.answers`. A first-try success is deterministic, but nothing pins that it happened.
4. **Reproducibility seals are inert** — most later rungs' `sha256sum -c` gates are commented out / `SEAL_FATAL=0`, `*.answers` are `# UNPINNED`, and the six rustc rungs don't declare `.answers` as a `Local` dep at all, so the seal check at `rustc-1.95.0/build.sh:604` is dead code. The PR body discloses pinning as a follow-up — this is just a request to track it somewhere concrete.
5. **`packages/rust` diverges from the chain's own claims** — unlike every rung it has no curl/wget/git stubs and no network tripwire, and `build.ncl` still declares `needs = { dns, internet }`, while the headline is "no recipe in the chain touches the network at build time." Either add the tripwire or state in the PR why the top-level package is deliberately exempt.
6. **Comment drift** — `rustc-1.94.1/build.ncl:3` says its stage0 is "the rustc-1.94.1 package" (should be 1.93.1); `binutils-2.46-glibc/build.ncl:3` says "same sha as packages/binutils" but that package is now 2.47; `glibc-bedrock-2.42/build.ncl` similarly vs glibc 2.44.
7. **Nits** — the "OutputBin's checker rejects the interpreter" rationale is inaccurate (`OutputBin` only special-cases shebang scripts, never ELF interpreters); the `find ./*/ -delete` clean-state blocks are safe only because every `Source` is `extract=false`; a few import identifiers don't match package names (`stage0-tcc-r2`, `stage0-cc`, `stage0-binutils-r10`, …).

The ladder machinery is solid and the central claims hold. Findings 1–2 are quick fixes and 3–5 are worth addressing or explicitly waiving before this lands. Not blocking on my side, but I didn't want it to merge on a green CI + bot-skip alone.

bryan-minimal:
Thanks for the careful read. Fixes for 1, 2, 4, 5, 6 and 7 are in b638f0a8; 3 and part of 4 are waived below with the reasoning.

1. **Dead arm64 Source arms** — removed the inner `match` in all three `-glibc` specs (the amd64 `Source` is now unconditional inside `amd64_spec`) and the unreachable `aarch64` blocks in their build.sh.
2. **s1 fallback** — an extract or `cd` failure now exits 1. Nothing is copied into `$OUTPUT_DIR` on that path any more.
3. **s1/s3 retry loops** — waived, with the reasoning on record: the perturbation loop exists because mes-m2's arena layout makes the first tcc compile a coin flip (an ASLR-shaped lottery in the seed compiler, not in our recipes; root-caused separately). The loop only re-runs the identical compile until it does not crash; a run that completes produces the same bytes. The rung pins for s1/s3 are in the same tracking item as 4.
4. **Seals inert** — the six rustc rungs now declare `rustc-<v>.answers` as a `Local` input, so the seal check in build.sh reads a real file instead of being dead code. The pins themselves stay `# UNPINNED` in this PR: cutting them needs a second cold build per rung on the merged recipes, which is the follow-up already listed in the description. It stays on the follow-ups list in the description.
5. **`packages/rust` tripwire** — added the same curl/wget/git stub harness the rungs use (PATH precedence asserted, `NETWORK-TRIPWIRE` checked after install) and dropped `needs = { dns, internet }`. The top-level package now makes the same claim as the chain.
6. **Comment drift** — fixed all three.
7. **Nits** — the `OutputBin` rationale now says why the binutils programs are `OutputData` (consumers reach them by absolute path under the package's own prefix); the stage0 imports are named after their packages instead of ladder-rung labels. The `find ./*/ -delete` blocks are indeed safe only because every `Source` is `extract = false`; that is the invariant those recipes rely on, and it is stated next to each.

Note for the merge: the s1 and `-glibc` build.sh changes move spec hashes at the bottom of the chain, so buildbot will run a cold chain build again after this push.

bryan-minimal:
Follow-up 318d00c7: the new `rust` tripwire fired on buildbot on `git submodule update --init curl` from the vendored curl-sys build.rs (the build itself had completed; the stub had refused the call). The rungs pre-empt that call with their P0c block, an empty `.git` marker in the bundled submodule dir; ported the same block to `packages/rust`.

bryan-minimal:
32d7df39 restores `needs = { dns, internet }` on `packages/rust`. Dropping it for finding 5 broke the Rust programs downstream: the builder grants network to a package when any of its *dependencies* declares `needs.internet` (minimal `crates/op/src/specs.rs`, the OR over dependencies), and ripgrep and friends fetch their crates during their own builds through that inherited flag (buildbot: `cargo build` → crates.io → TLS verify failed without the sandbox network setup). rust's own build stays offline and tripwired; the comment on the flag now says why it is there.

mitodrummer (approved):

PR #691, opened 2026-09-11
helix-grammars: read the version from helix, not the first grammar repo (pkgmgr#773)

bryan-minimal:
helix-grammars carries no `source_provenance` on purpose (it aggregates 43 grammar repos), so pkgmgr's version check fell back to the **first build_dep's** source — `tree-sitter/tree-sitter-bash`, v0.25.1 — and compared that against the package's helix-scheme version 25.07.1. The no-downgrade clamp rejected it every day and the package rendered as plain **"up to date"** (pkgmgr#773): a real helix release would have been invisible.

The version *is* helix's (25.07.1 = the helix release whose grammar pin-set this is), so the check now reads helix's latest release via `version_check_url` + `version_check_regex`. The regex deliberately contains no quotes or backslashes — they don't survive the Nickel → pkgmgr round trip (first attempt arrived as a lone `\`); `.` stands in for the JSON quote characters.

**Verified locally** with `pkgmgr check --package helix-grammars`: the custom tier resolves helix and the package reports up to date at 25.07.1 instead of a degraded tier / daily clamp.

Attrs-only; nothing about the build moves.

## Summary by CodeRabbit

* **Chores**
  * Updated Helix release version checks to use the latest published release information.
  * Improved semantic version extraction for more reliable version detection.

### DIFF
--- a/packages/helix-grammars/build.ncl
+++ b/packages/helix-grammars/build.ncl
@@ -315,5 +315,17 @@ let version = "25.07.1" in
       license_spdx = "MIT AND Apache-2.0 AND WTFPL",
       # source_provenance is deliberately absent: there is no single upstream to
       # point at, the package aggregates the 43 repositories pinned above.
+      #
+      # The VERSION, though, is helix's (25.07.1 = the helix release whose
+      # grammar pin-set this is), so the version check must read helix — not
+      # the first grammar repo. Without this, pkgmgr fell back to the first
+      # build_dep's source (tree-sitter/tree-sitter-bash, v0.25.1), compared it
+      # against 25.07.1 and clamped it as a "downgrade" every day (pkgmgr#773):
+      # a real helix release would have been invisible.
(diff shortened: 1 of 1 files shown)


bryan-minimal:
Closing: `Attrs` is a **closed** contract in minimal's stdlib (`attr_classes.ncl`), and `version_check_url` / `version_check_regex` have never been declared there — hence the 2-second `failed to load graph for amd64`. The pkgmgr escape hatch is unreachable from a build.ncl (no package in pkgs history has ever set it).

Fixed without a pkgs change instead:
- minimal-supply-chain#432 adds a package-name override arm: `helix-grammars` → highest stable tag of `helix-editor/helix`.
- pkgmgr-rs#810 makes the resolver cascade continue past a tier that answers *older than what we ship* (the derived tree-sitter-bash tier), so that override arm is actually reached.

Verified live: helix-grammars resolves 25.07.1 → up to date, no daily clamp.

PR #693, opened 2026-09-12
lean: ship the .olean.server/.olean.private sidecars — lean still compiled nothing after #605

bryan-minimal:
pkgs#605 shipped the module roots' `.olean`, and `lean` **still** cannot compile `#eval 1+1` (verified in a fresh session built from main today):

```
error: failed to open file '/usr/lib/lean/Init.olean.server': No such file or directory
```

Since Lean 4.2x every module is **three** files — `X.olean`, `X.olean.server`, `X.olean.private` — and lean refuses to load a module whose sidecars are missing. Both the install script (`cp *.olean`) and the output globs (`**/*.olean`, `*.olean`) carried only the first: the session's tree has **2,485 `.olean` and 0 sidecars**.

## Change
- `build.sh`: `cp *.olean*` (+ `*.ilean`, the language-server index) at the root; the root-module check now also demands `$m.olean.server`, so the next layout change fails the build instead of shipping a lean that compiles nothing.
- `build.ncl`: `**/*.olean*` / `*.olean*` globs, plus `.ilean` outputs.
- **New standalone test**: compiles `#eval 1+1` and `theorem t : 1 + 1 = 2 := rfl`. This exact check has now failed twice for two different missing-file classes; it belongs in the package.

## Why it matters
This is the last step of the aeneas verification bench (charon → aeneas → **lean**). The first two have worked since 2026-08-13; with this, `./verify` in the lab can actually check a model.

## Summary by CodeRabbit

- **New Features**
  - Added the CaDiCaL 2.1.2 SAT solver package, including its command-line executable and runtime support.
  - Lean packages now include compiled module files, `.olean` sidecars, `.ilean` indexes, and intermediate representation artifacts.
  - Root-level compiled artifacts are included.
  - Lean runtime packaging includes required OpenSSL support.

- **Bug Fixes**
  - Lean installations now verify required server sidecars and report an error when one is missing.

- **Tests**
  - Added coverage for CaDiCaL version reporting and SAT formula solving.
  - Added validation for Lean compilation, Lake configuration, and builds.

### DIFF
--- a/packages/cadical/build.ncl
+++ b/packages/cadical/build.ncl
@@ -1,6 +1,7 @@
 let { subsetOf, Attrs, BuildSpec, Local, OutputBin, Source, Test, .. } = import "minimal.ncl" in
 let base = import "../base/build.ncl" in
 let toolchain = import "../toolchain/build.ncl" in
+let make = import "../make/build.ncl" in
 let gcc = import "../gcc/build.ncl" in
 let glibc = import "../glibc/build.ncl" in
 
@@ -22,6 +23,7 @@ let version = "2.1.2" in
     } | Source,
     base,
     toolchain,
+    make,
   ],
 
   runtime_deps = [
--- a/packages/cadical/build.ncl
+++ b/packages/cadical/build.ncl
@@ -1,0 +1,64 @@
+let { subsetOf, Attrs, BuildSpec, Local, OutputBin, Source, Test, .. } = import "minimal.ncl" in
+let base = import "../base/build.ncl" in
+let toolchain = import "../toolchain/build.ncl" in
+let gcc = import "../gcc/build.ncl" in
+let glibc = import "../glibc/build.ncl" in
+
+# The release Lean 4.33 bundles as `bin/cadical` (lean4 CMake INSTALL_CADICAL).
+let version = "2.1.2" in
+{
+  name = "cadical",
+  # CaDiCaL, the SAT solver behind Lean's `bv_decide`. Official Lean releases
+  # ship it next to `lean`; a from-source lean build does not, and `bv_decide`
+  # then fails with "could not execute external process 'cadical'". It is found
+  # by name on PATH, so this package is enough.
+  build_deps = [
+    { file = "build.sh" } | Local,
+    {
+      url = "https://github.com/arminbiere/cadical/archive/refs/tags/rel-%{version}.tar.gz",
+      sha256 = "292c2bb8d712d6d05fce3d3df63b922b8fa45e03974a79f7bae5bf68c284f131",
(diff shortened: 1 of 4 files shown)


bryan-minimal:
Second commit, found by the new test rather than by reading: **`lean` links `libssl.so.3` / `libcrypto.so.3` and `openssl` was not a runtime dep.** In the standalone-test rootfs (base + declared runtime deps) `lean --version` exits 127 from the loader before printing anything. Every session that had ever run lean carried openssl through some other package, which is how it hid. `readelf -d /usr/bin/lean` for the record:

```
libstdc++.so.6 libInit_shared.so libleanshared_2.so libleanshared_1.so libleanshared.so libgmp.so.10 libuv.so.1 libssl.so.3 libcrypto.so.3 libm.so.6 libc.so.6
```

Worth noting for the `missing runtime_deps` checker: it passed on this package both before and after — a real DT_NEEDED gap it did not see.

Validated in a sandbox session: the first CI failure was this test with exit 127 on `/bin/lean --version`; with `openssl` declared the package is rebuilding locally now and CI is rebuilding here.

bryan-minimal:
Third commit, from building the first package that uses `lake` (aeneas-latest): **the per-module compiled IR (`X.ir` + `X.ir.sig`) was never installed either** — 2,483 files in the reference toolchain, 0 in ours. `lean` elaborates and `#eval`s without them (so the compile test passed), but `lake` compiles the definitions in a lakefile and fails its `compiler IR check` on the first private `Init` declaration they touch:

```
error: lakefile.lean:12:18: failed to compile definition, compiler IR check failed at `config._closed_5`. Error: depends on declaration '_private.Init.Data.Array.Basic…'
```

So until this commit no Lake project could be configured with our lean at all. Added to the copies and globs, the root-module check demands `Init.ir`, and a new standalone test configures and builds a minimal Lake project (`LAKE_NO_CACHE=1`), which is the exact thing that was broken. Three file classes missing from one package, each only visible once the previous one was fixed: roots (#605), sidecars, IR.

bryan-minimal:
Fourth commit: **`bv_decide` needs the `cadical` SAT solver on PATH**, which official Lean releases bundle in `bin/` (2.1.2 for 4.33) and a from-source build does not produce — `could not execute external process 'cadical'` (found building aeneas-latest, whose Bvify module exercises `bv_decide` in its own `#guard_msgs` examples). Adds a small `cadical` package (MIT, rel-2.1.2, `./configure && make`, SAT smoke test) and makes it a runtime dep of `lean`. That's the fourth bundled piece missing from this package: roots, sidecars, IR, cadical.

mitodrummer (approved):

PR #699, opened 2026-09-15
unzip: move to Debian's 6.0-31 patch series (3 memory-safety fixes)

bryan-minimal:
## What

We were shipping Debian's **6.0-29** unzip patch series. Debian is on **6.0-31**. The two revisions we skipped carry three memory-safety fixes, all reachable from attacker-controlled archive bytes.

| Fix | Debian rev | Bug | Class | Files |
|---|---|---|---|---|
| CAN-2026-2034440 | 6.0-30 | [#1142904](https://bugs.debian.org/1142904) | heap out-of-bounds **read**, `EF_IZUNIX3` extra-field handler | `process.c`, `unzpriv.h` |
| CAN-2026-2034443 | 6.0-30 | [#1142905](https://bugs.debian.org/1142905) | stack out-of-bounds NUL **write**, `EF_SMARTZIP` handler | `zipinfo.c` |
| CAN-2026-2034442 | 6.0-31 | [#1142906](https://bugs.debian.org/1142906) | heap buffer overflow **write**, `memextract()` STORED path | `extract.c` |

All three are from Paul Marquess, carried by Debian. Info-ZIP unzip 6.0 is frozen upstream (2009), so the distro patch series is the only place its security fixes exist.

## Two changes beyond the bump

**The overlay is now Debian's tarball mirrored verbatim.** Previously we vendored a hand-repacked archive with renumbered patch filenames, which meant the pinned sha256 was only ever checkable against itself. Now the pin is the hash Debian publishes in their PGP-signed `unzip_6.0-31.dsc`, so a reviewer can verify it without trusting our repack:

```
$ curl -s https://deb.debian.org/debian/pool/main/u/unzip/unzip_6.0-31.dsc | grep debian.tar
 d7dc23f5254c6711dc05f467e70236cfd50a7931001c6152904474b19f295c38 27232 unzip_6.0-31.debian.tar.xz
```

That is the value in `build.ncl`. I also re-downloaded from a second Debian mirror and re-hashed the object after the GCS round trip; both match. Patch paths move to `debian/patches/`.

**`build.sh` now fails closed on a short series.** It counts what it applied and refuses to build if the count is not 34. The failure this guards is the quiet one: a truncated or swapped overlay produces an unpatched `unzip` that passes every test we have, because none of our tests feed it a malicious archive. A silently unpatched zip extractor is worse than a failed build.

## What did not change

The base `unzip_6.0.orig.tar.gz` is byte-identical between revisions, so it is untouched. Of the 31 patches we already carried, 30 are byte-identical to Debian's current copies. `cve-2015-7696.patch` differs only by a dropped `Forwarded: yes` metadata header.

## How this got missed

This is the part worth more than the patch, and I would not want it to merge unremarked.

Nothing watches Debian revisions. `pkgmgr check` asks upstream for the latest version, upstream says `6.0`, we match, and the package reports **current** forever while the patch series rots underneath it. We sat on 6.0-29 for three months. The package was never "update-dark" by any signal we collect, because by our own definition it was up to date.

This PR adds a comment on the source saying so, with the command to diff against Debian. That is a sticky note, not a fix. The real fix is for the version checker to understand distro-patch provenance for frozen-upstream packages, and it needs its own issue. **unzip is unlikely to be the only package in this shape** — anything where we vendor a distro patch series against a dead upstream has the same blind spot.

## Verification

Built in a `pkgbuild` sandbox on arm64, not locally:

- `min package build --rebuild --verbose unzip` — exit 0, all 34 patches applied, no failed hunks or fuzz. The three new patches touch exactly the file sets listed in the table above.
- `min check --packages unzip` — all 15 checks pass, including `standalone tests` (the round-trip extraction test) and `enumerate bins`.

## Review notes

I have not tested against a PoC archive for any of the three issues; the evidence that they are fixed is that Debian's patches apply cleanly and completely. If we want stronger assurance than "the distro's fix is in the binary", that is a fuzzing corpus, not a review comment.

Needs a human reviewer per repo policy.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the unzip package to incorporate Debian’s latest 6.0-31 security patch series.
  - Included additional CVE protections and build fixes.
  - Improved source verification and patch integrity checks to detect incomplete or unexpected patch updates.
  - Preserved compatibility fixes for obsolete time-function declarations.

- **Documentation**
  - Documented the effective security version, source-verification reference, and limitations around automatically detecting Debian revision updates.


bryan-minimal:
## What this does not fix: CVE-2021-4217

Checked while reviewing the series, so a reviewer does not have to wonder.

**CVE-2021-4217 remains unfixed by 6.0-31**, as it was by 6.0-29. It is a NULL pointer dereference in `getUnicodeData()` in `process.c`: a zip entry carrying an `EF_UNIPATH` (0x7075) extra field but no filename leaves `G.filename_full` NULL, which goes straight to `strlen()`. The same upstream fix also covers a short-read in `do_string()` in `fileio.c` that leaves the extra-field buffer partly uninitialized.

No patch in the 34-patch 6.0-31 series touches it. The only `filename_full` reference in the whole series is an unrelated line inside `cve-2022-0529-and-cve-2022-0530.patch`.

This is a deliberate distro split, not an oversight on our part:

- **Debian declined it.** Every suite is marked `vulnerable`, fixed version `(unfixed)`, urgency `unimportant`, with the note "Crash in CLI tool, no security impact".
- **Ubuntu patched it** (`debian/patches/CVE-2021-4217.patch`, Nils Bars, 2022-01-17) with a NULL guard returning `PK_ERR`.
- **Red Hat deferred it** (bugzilla 2044583, still NEW; RHEL 8/9 `Fix deferred`).

I am not adding it to this PR. Mixing an Ubuntu patch into a Debian series is how a vendored patch set starts drifting from any upstream we can diff against, and that drift is the thing that let this package fall two revisions behind in the first place (pkgmgr-rs#820). If we want it, it should be a separate, explicitly-labelled local patch.

Severity supports leaving it: CVSS 3.3, local, requires user interaction, availability-only. The "code execution" phrasing in the record is the reporter's speculation, not a demonstrated outcome. Our `unzip` is a build-time dependency of `bun` and the two chromium binary packages, so the exposure is a build-time crash on a hostile archive, and our build inputs are hash-pinned.

The three fixes this PR **does** add are all memory-safety defects with write primitives among them, which is a different class from a NULL-deref crash.

mitodrummer (approved):

PR #711, opened 2026-09-17
vim 9.2.1116, gh 2.100.0, broot 1.60.1 (3 advisories)

bryan-minimal:
## What

Three of the fleet's remaining non-chromium advisories, each cleared by a routine version bump.

| package | bump | advisory | result |
|---|---|---|---|
| vim | 9.2.1036 → 9.2.1116 | `GHSA-x9jf-rjm6-vxqh` (LOW) | 1 → **0** |
| gh | 2.97.0 → 2.101.0 | `GHSA-vfhh-p7hm-pxfh` (LOW) | 1 → **0** |
| broot | 1.58.0 → 1.60.1 | `CVE-2026-72847` (LOW) | 1 → **0** |

Verified by running the scanner against both trees, not by assuming a newer version helps. Artifacts confirmed mirrored before opening.

## This started as six packages

Three were dropped after measuring, and two of them turned out to be **unfixable by any bump** — which is the more useful finding.

### harfbuzz 14.3.1 → 14.4.0 does not clear its advisory

```
GHSA-f6q6-4vjq-m6j7 (LOW)
  STILL AFFECTING — no fix metadata to compare against (fix_status: unknown_fix).
  No fixed version, no affected range, no fix commit.
```

There is nothing to compare against, so **no version will ever clear it**. Bumping would have rebuilt both chromium packages, ffmpeg, libass and pango for zero security gain.

### kubectl 1.36.3 → 1.37.0 does not clear its advisory

```
GO-2025-3547 (LOW)
  STILL AFFECTING — 1.37.0 falls inside an affected range: [>= 1.3.0]
  affected_ranges=[>= 1.3.0]   aliases=CVE-2024-7598, GHSA-r56h-j38w-hrqq
```

**Unbounded above.** Every version from 1.3.0 onward is "affected", permanently — the same shape as the glibc false positives cleared in minimal-supply-chain#435.

Both are **`advisory_overrides` candidates, not update candidates**. A finding that no action can clear is not a finding; it is noise that trains people to ignore the list. I'd rather route them there than keep bumping packages that cannot be helped.

### croc was reported updated but nothing was written

The updater printed `✓ croc: updated build.ncl` and logged `updated=6`, but only **5 files changed** — `packages/croc/build.ncl` is untouched at 11.1.0. Filed separately; it is the ship-half-and-say-shipped shape, and the summary counter agreeing with the lie is what makes it dangerous.

## Context

From a full-fleet scan: **563 still-affecting advisories across 38 packages**, of which **460 (82%) are the chromium pair** (the settled arm64 cliff — not bump-fixable). Of the 36 non-chromium packages, only 10 had any newer version at all; the rest are already at latest upstream. So this PR is most of what a version bump can still buy on the leaf packages.

## Summary by CodeRabbit

- **Updates**
  - Updated broot to version 1.60.1.
  - Updated GitHub CLI (`gh`) to version 2.101.0.
  - Updated Vim to version 9.2.1116.


bryan-minimal:
## CI caught a real break — gh retargeted to 2.100.0

```
go: go.mod requires go >= 1.27.0 (running go 1.26.6; GOTOOLCHAIN=local)
```

gh moved its `go.mod` floor between releases:

| gh | go floor | |
|---|---|---|
| 2.97.0 (shipped) | 1.26.0 | |
| **2.100.0** | **1.26.0** | newest that builds on our toolchain |
| 2.101.0 | 1.27.0 | what the updater offered |

Retargeted to **2.100.0**, which still clears `GHSA-vfhh-p7hm-pxfh` — scanned at 2.100.0, `total_vulns=0`. Security outcome unchanged, toolchain untouched.

Bumping Go to 1.27 to take 2.101.0 is a toolchain change with its own cascade and doesn't belong in a leaf-package vuln PR.

**Worth noting for the updater:** it offered a version the fleet cannot compile. Nothing consults the upstream's own toolchain floor (go.mod `go`, `rust-version`, `engines.node`) against what we ship, so this class fails at CI rather than at selection. Will file separately.

mitodrummer (approved):

PR #715, opened 2026-09-17
biome: 2.5.8 -> 2.5.14, and stamp the version into the binary

### LINKED ISSUE #637: biome package reports Version: 0.0.0 — version not stamped at build

mitodrummer:
Bumps `biome` 2.5.8 → 2.5.14 and fixes the version stamping that made `biome --version` report `Version: 0.0.0`.

Closes #637. Supersedes #650.

## The bug

`biome_cli`'s Cargo version is literally `0.0.0` in the source tree — upstream stamps the real version at release time through a `BIOME_VERSION` environment variable, read at compile time in two places:

```rust
// crates/biome_cli/src/lib.rs
pub(crate) const VERSION: &str = match option_env!("BIOME_VERSION") {
    Some(version) => version,
    None => env!("CARGO_PKG_VERSION"),   // "0.0.0"
};

// crates/biome_configuration/src/lib.rs
pub const VERSION: &str = match option_env!("BIOME_VERSION") {
    Some(version) => version,
    None => "0.0.0",
};
```

Our build never set it, so the binary reported `0.0.0`. The user-visible consequence is the configuration one: no `$schema` value can ever match `0.0.0`, so biome emits a "configuration schema version does not match the CLI version" notice on every run against any config that sets `$schema` — permanent cosmetic noise for every consumer. It surfaced while migrating gominimal/webapp off the npm `@biomejs/biome` devDependency onto the catalog package.

## The fix

- Forward `version` through `build_args` and `export BIOME_VERSION="$MINIMAL_ARG_VERSION"` in `build.sh`, with a comment recording why it is load-bearing so a future bump doesn't drop it.
- Pin it with a standalone test asserting the **exact** `--version` output. Deliberately exact: a substring match would still pass against `0.0.0`.
- Bump to 2.5.14 (2026-09-16), the current release. `strip_prefix` pattern is unchanged; `sha256` verified against the upstream archive over two independent downloads.

## On #650

An outside contributor, @npetrovcode, diagnosed this and submitted the same `BIOME_VERSION` fix in #650 on 2026-08-26. It could not be merged: the CLA went unsigned and they have not been reachable since. Their approach was correct and is kept here — `build_args`, the `BIOME_VERSION` export, and the exact-match regression test are all theirs. This PR rebases that onto current main, adds the version bump and the explanatory comments, and runs the validation their environment couldn't. #650 is closed with credit and an offer to hand it back if they sign.

## Verification

- `min check` — all checkers pass for `biome`.
- `min package patched-build biome` — succeeds; the built binary reports the stamped version, which is what the new test asserts.

https://claude.ai/code/session_01V5zRHukXJu6dNykVTjhHjV

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Biome to version 2.5.14.
  * Ensured the reported Biome version matches the packaged version.
  * Added regression coverage to verify accurate version reporting.


mitodrummer:
Build verification, now complete (the PR body was written while this was still compiling):

```
min package patched-build biome   # Finished `release` profile in 11m 37s, exit 0
min check                         # re-run post-build
```

All 15 checkers pass for `biome`, and the four that report `Skip` on an unbuilt package now run for real:

| Checker | Before build | After build |
|---|---|---|
| `enumerate bins` | Skip | **Pass** |
| `output types valid` | Skip | **Pass** |
| `missing runtime_deps` | Skip | **Pass** |
| `standalone tests` | Skip | **Pass** |

`standalone tests` passing is the one that matters here: it executes `version_is_stamped`, which asserts `biome --version` is exactly `Version: 2.5.14`. That test fails against an unstamped binary, so this is a direct confirmation the `BIOME_VERSION` export lands.

bryan-minimal (approved):

